PR #4, opened 2015-07-10
added some more parameters for metadata transformation

dweidenfeld:
added
- Last-Modified
- Display-Url
- Crawl-Once
- Lock

as parameters to be passed through to the metadata/document transformer, so that you can modify metadata based on more information

### DIFF
--- a/src/com/google/enterprise/adaptor/LoggingFilter.java
+++ b/src/com/google/enterprise/adaptor/LoggingFilter.java
@@ -37,6 +37,7 @@ class LoggingFilter extends Filter {
 
   @Override
   public void doFilter(HttpExchange ex, Filter.Chain chain) throws IOException {
+    long startMillis = System.currentTimeMillis();
     try {
       log.fine("beginning");
       logRequest(ex);
@@ -51,6 +52,7 @@ class LoggingFilter extends Filter {
       throw e;
     } finally {
       logResponse(ex);
+      logDurationInMillis(ex, startMillis);
       log.fine("ending");
     }
   }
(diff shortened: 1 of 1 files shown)


dweidenfeld:
(1) is fixed

(2) is fixed with as good as I could do it. Is there some rule how to group your imports?
I installed the google-code-style.xml in my IntelliJ, so I thought that this should work for you ;)

BrettMichaelJohnson:
On Jul 16, 2015, at 12:50 AM, Dominik Weidenfeld notifications@github.com wrote:

> (1) is fixed
> 
> (2) is fixed with as good as I could do it. Is there some rule how to group your imports?

The rules for grouping and sorting imports are outlined
in the Google Java style guide: 
https://google-styleguide.googlecode.com/svn/trunk/javaguide.html#s3.3-import-statements

The import statements were in the correct order and
spacing in the original file, and you neither added nor

removed any.  So simply restoring the imports to their
original arrangement should suffice.

Your pull request also contains numerous changes in 
whitespace with no other functional code changes.
Some were changes in indentation, but most were
deleting trailing whitespace on lines.
This is discouraged for several reasons.  It makes
it difficult for us as reviewers to focus on your 
actual functional changes.  It also pollutes the 
version control history for those lines.

When problems are found with a particular piece of
code, we often use git blame to determine the author
and commit revision where that code was introduced
or modified.  We then examine all the changes for that
particular commit, the log entry, and the code review,
to try to understand why the code was written as it
was.  Gratuitous changes alter that history.

Since more than half of the changes in the code
are non-functional changes (re-ordering the imports
and changing whitespace), you may find it more
convenient to use git reset to restore the files to their
original state, then re-apply just your changes to 
the transform() method.

—
Brett M. Johnson

PapaJoltOmega:
Thanks Brett for the nice explanation.

I've added a couple of comments to the 2nd commit.

Thank you for this work,
PJ

dweidenfeld:
Thank you for your response. I will change the code style and implement the checks as you wanted (maybe beginning next week).
The code style changes has only been made because the IDE optimises the code on save automatically. I will try to revert my changes completely and only add my changes (maybe with a normal text editor).

PR #5, opened 2015-07-13
content transformer

dweidenfeld:
added the possibility to change the content from each adaptor by an external transformer. the function is nearly equal to the metadata transformer but handles the content.

### DIFF
--- a/src/com/google/enterprise/adaptor/LoggingFilter.java
+++ b/src/com/google/enterprise/adaptor/LoggingFilter.java
@@ -37,6 +37,7 @@ class LoggingFilter extends Filter {
 
   @Override
   public void doFilter(HttpExchange ex, Filter.Chain chain) throws IOException {
+    long startMillis = System.currentTimeMillis();
     try {
       log.fine("beginning");
       logRequest(ex);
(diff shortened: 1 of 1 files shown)


PapaJoltOmega:
Thank you for your work.
I've left a couple of comments regarding following Google's style guide for Java source code.

Further, could you please add an example into the adaptor/examples directory?  Something simple like converting lower case to upper case.  I'd like to see better how this approach works in practice.  

Thank you,
PJ

src/com/google/enterprise/adaptor/DocumentHandler.java
```diff
@@ -687,6 +690,12 @@ public OutputStream getOutputStream() throws IOException {
             ex.getResponseBody()));
         os = countingOs;
       }
+      if (null != contentTransformer) {
```
ejona86: This does have some threading issues, as we will be handling multiple requests at the same time. We're going to need a way to create multiple of the content transformers. Probably create a new one for each request.

BrettMichaelJohnson: Actually, I was thinking we might want to create multiple content transforms
per request in a transform pipeline like the metadata transforms.
—
Brett M. Johnson

dweidenfeld:
I will close this pull request and do another one with the expected behaviour, so that the merge history will get clean.

ejona86:
Note that if you do a forced push to your branch, the pull request will be updated with the new history. You can certainly choose to make a new pull request, but you don't need to in order to clean up the git history.

PR #6, opened 2015-07-20
added more parameters for metadata transformation

dweidenfeld:
I added it as new pull request, to get a clean merge history.
(Original: https://github.com/googlegsa/library/pull/4)

### DIFF
--- a/src/com/google/enterprise/adaptor/DocumentHandler.java
+++ b/src/com/google/enterprise/adaptor/DocumentHandler.java
@@ -1000,8 +1000,34 @@ class DocumentHandler implements HttpHandler {
       Map<String, String> params = new HashMap<String, String>();
       params.put("DocId", docId.getUniqueId());
       params.put("Content-Type", contentType);
+      if (null != lastModified) {
+        params.put("Last-Modified", "" + lastModified.getTime());
+      }
+      if (null != displayUrl) {
+        params.put("Display-Url", "" + displayUrl);
+      }
+      params.put("Crawl-Once", "" + crawlOnce);
+      params.put("Lock", "" + lock);
       transform.transform(metadata, params);
       contentType = params.get("Content-Type");
+      try {
+        final long lm = Long.parseLong(params.get("Last-Modified"));
+        if (null != params.get("Last-Modified") && lastModified.getTime() != lm) {
+          lastModified = new Date(lm);
+        }
+      } catch (NumberFormatException e) {
+        log.finer("Cannot change back the last-modified date");
+      }
+      try {
+        final String du = params.get("Display-Url");
+        if (null != du) {
+          displayUrl = new URI(du);
+        }
+      } catch (URISyntaxException e) {
+        log.finer("Cannot create display-url");
+      }
+      crawlOnce = Boolean.parseBoolean(params.get("Crawl-Once"));
+      lock = Boolean.parseBoolean(params.get("Lock"));
     }
 
     private class CloseNotifyOutputStream extends FastFilterOutputStream {
--- a/src/com/google/enterprise/adaptor/DocumentHandler.java
+++ b/src/com/google/enterprise/adaptor/DocumentHandler.java
@@ -1022,8 +1022,9 @@ class DocumentHandler implements HttpHandler {
           }
         }
       } catch (NumberFormatException e) {
-        log.log(Level.FINER,
-            "Failed changing last-modified date {0}", params.get("Last-Modified-Millis-UTC"));
+        log.log(Level.WARNING,
+            "Failed changing last-modified date {0}",
+            params.get("Last-Modified-Millis-UTC"));
       }
       try {
         final String du = params.get("Display-URL");
@@ -1031,7 +1032,8 @@ class DocumentHandler implements HttpHandler {
           displayUrl = new URI(du);
         }
       } catch (URISyntaxException e) {
-        log.log(Level.FINER, "Failed changing display URL {0}", params.get("Display-URL"));
+        log.log(Level.WARNING, "Failed changing display URL {0}",
+            params.get("Display-URL"));
       }
       crawlOnce = Boolean.parseBoolean(params.get("Crawl-Once"));
       lock = Boolean.parseBoolean(params.get("Lock"));
--- a/src/com/google/enterprise/adaptor/DocumentHandler.java
+++ b/src/com/google/enterprise/adaptor/DocumentHandler.java
@@ -17,6 +17,7 @@ package com.google.enterprise.adaptor;
 import static java.util.Map.Entry;
 
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Strings;
 
 import com.sun.net.httpserver.HttpExchange;
 import com.sun.net.httpserver.HttpHandler;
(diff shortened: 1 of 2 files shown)


PapaJoltOmega:
Thank you.  I've added some comments to the PR.

FYI: I'll be away on holiday for next two weeks.

Thank you,
PJ

dweidenfeld:
I have made the wanted changes and added some tests to be sure everything works as expected.

> save the displayUrl string passed in, compare using equals on strings, if param didn't change then avoid call to new URI and setting

This is the only thing I have not changed, because it could be that the Display-URL is set to null (currently allowed) and if the transformer sets the URL from null to something else the comparison is not necessary. The performance should also not be affected.

src/com/google/enterprise/adaptor/DocumentHandler.java
```diff
@@ -1000,8 +1001,40 @@ private void transform() {
...
+          }
+        }
+      } catch (NumberFormatException e) {
+        log.log(Level.FINER,
+            "Failed changing last-modified date {0}", params.get("Last-Modified-Millis-UTC"));
+      }
+      try {
+        final String du = params.get("Display-URL");
+        if (!Strings.isNullOrEmpty(du)) {
+          displayUrl = new URI(du);
```
PapaJoltOmega, line 1032: At this point we know that du is not null, so we could safely:

if (!du.equals(origDisplayUrl)) {
  displayUrl = new URI(du);
}

In most cases du will equal origDisplayUrl and we can then in most cases skip all the parsing that URI does to get all the various parts it gets (host, port, path, query, etc.).

Also, consider changing origDisplayUrl to origDisplayUrlStr.

PapaJoltOmega:
Great.  Thank you.

One comment regarding the displayUrl changed check. And one comment regarding slight difference in treatment of null lastModified and null displayUrl.

Thank you very much,
PJ

src/com/google/enterprise/adaptor/DocumentHandler.java
```diff
@@ -1000,8 +1001,40 @@ private void transform() {
...
       params.put("Content-Type", contentType);
+      if (null != lastModified) {
+        params.put("Last-Modified-Millis-UTC", "" + lastModified.getTime());
+      }
+      String origDisplayUrl = null;
+      if (null != displayUrl) {
+        origDisplayUrl = "" + displayUrl;
+      }
+      params.put("Display-URL", origDisplayUrl);
+      params.put("Crawl-Once", "" + crawlOnce);
```
PapaJoltOmega, line 1012: for lastModified we don't include the key at all if it is null

for displayUrl we explicitly include null if it is null

why not skip the params.put("Display-URL", origDisplayUrl) if displayUrl is null?

dweidenfeld, line 1012: Okay, done. Nice idea ;)

src/com/google/enterprise/adaptor/DocumentHandler.java
dweidenfeld, line 1032: I added the equality check. (I did not remove the null, check because if the value will be null and you call .equal on null (you know what will happen) ;)
Also the JVM will optimize the routines regarding the null checks on multiple runs.

src/com/google/enterprise/adaptor/DocumentHandler.java
```diff
@@ -1000,8 +1001,40 @@ private void transform() {
...
+        final String lmMillisUTC = params.get("Last-Modified-Millis-UTC");
+        if (!Strings.isNullOrEmpty(lmMillisUTC)) {
+          final long lm = Long.parseLong(lmMillisUTC);
+          if (lastModified.getTime() != lm) {
+            lastModified = new Date(lm);
+          }
+        }
+      } catch (NumberFormatException e) {
+        log.log(Level.FINER,
+            "Failed changing last-modified date {0}", params.get("Last-Modified-Millis-UTC"));
```
PapaJoltOmega: over80

src/com/google/enterprise/adaptor/DocumentHandler.java
```diff
@@ -1000,8 +1001,40 @@ private void transform() {
...
+      try {
+        final String lmMillisUTC = params.get("Last-Modified-Millis-UTC");
+        if (!Strings.isNullOrEmpty(lmMillisUTC)) {
+          final long lm = Long.parseLong(lmMillisUTC);
+          if (lastModified.getTime() != lm) {
+            lastModified = new Date(lm);
+          }
+        }
+      } catch (NumberFormatException e) {
+        log.log(Level.FINER,
```
PapaJoltOmega: make into warning as likely there is a bug in transformer

src/com/google/enterprise/adaptor/DocumentHandler.java
```diff
@@ -1000,8 +1001,40 @@ private void transform() {
...
+            "Failed changing last-modified date {0}", params.get("Last-Modified-Millis-UTC"));
+      }
+      try {
+        final String du = params.get("Display-URL");
+        if (!Strings.isNullOrEmpty(du) && !du.equals(origDisplayUrlStr)) {
+          displayUrl = new URI(du);
+        }
+      } catch (URISyntaxException e) {
+        log.log(Level.FINER, "Failed changing display URL {0}", params.get("Display-URL"));
```
PapaJoltOmega: over80

test/com/google/enterprise/adaptor/DocumentHandlerTest.java
```diff
@@ -735,6 +753,72 @@ public void getDocContent(Request request, Response response)
...
+    String remoteIp = ex.getRemoteAddress().getAddress().getHostAddress();
+    DocumentHandler handler = createHandlerBuilder()
+        .setAdaptor(adaptor)
+        .setFullAccessHosts(new String[]{remoteIp, "someUnknownHost!@#$"})
+        .setSendDocControls(true)
+        .setGsaVersion("7.4.0-0")
+        .setTransform(new TransformPipeline(Lists.newArrayList(new DocumentTransform() {
+          @Override
+          public void transform(Metadata metadata, Map<String, String> params) {
+            assertEquals(oDate.getTime(), Long.parseLong(params.get("Last-Modified-Millis-UTC")));
```
PapaJoltOmega: files that are previously 80 char maintain that limit.  new files can use 100 char limit, per style guide. i think this file is historically 80 char and so should maintain being 80 char per line.

PapaJoltOmega:
LGTM.  Thank you. 

I'm ready to merge this CL, but for nit comments.

Thank you,
PJ

src/com/google/enterprise/adaptor/DocumentHandler.java
dweidenfeld: fixed.

src/com/google/enterprise/adaptor/DocumentHandler.java
dweidenfeld: done.

src/com/google/enterprise/adaptor/DocumentHandler.java
dweidenfeld: fixed.

test/com/google/enterprise/adaptor/DocumentHandlerTest.java
dweidenfeld: fixed.

PapaJoltOmega:
LGTM.

Thank you.

PR #7, opened 2015-07-20
Added possibility to change the document content

dweidenfeld:
Improved version of https://github.com/googlegsa/library/pull/5

### DIFF
--- a/src/com/google/enterprise/adaptor/Config.java
+++ b/src/com/google/enterprise/adaptor/Config.java
@@ -174,7 +174,7 @@ import java.util.logging.Logger;
  *     and modify principals as described. Defaults no modifications
  * <tr><td> </td><td>transform.pipeline </td><td> sequence of
  *     transformation steps.  Defaults to no-pipeline
- * <tr><td> </td><td>content.transformer.pipeline </td><td>
+ * <tr><td> </td><td>content.transform.pipeline </td><td>
  *     content transformer pipeline. Defaults to no-pipeline</td></tr>
  * <tr><td> </td><td>saml.idpExpirationMillis </td><td> Expiration time
  *     sent in SAML Authentication response. Defaults to 30,000 milliseconds.
@@ -288,7 +288,7 @@ public class Config {
     addKey("adaptor.docContentTimeoutSecs", "180");
     addKey("adaptor.docHeaderTimeoutSecs", "30");
     addKey("transform.pipeline", "");
-    addKey("content.transformer.pipeline", "");
+    addKey("content.transform.pipeline", "");
     addKey("journal.reducedMem", "true");
     addKey("gsa.acceptsDocControlsHeader", "GENERATE", new ValueComputer() {
           public String compute(String rawValue) {
--- a/src/com/google/enterprise/adaptor/Config.java
+++ b/src/com/google/enterprise/adaptor/Config.java
@@ -597,7 +597,7 @@ public class Config {
   }
 
   /**
-   * Returns a map configuration the content transformer.
+   * Returns a map configuration the content transform pipeline.
    */
   List<Map<String, String>> getContentTransformFactorySpec() {
     return getListOfConfigs("content.transform.pipeline");
--- a/src/com/google/enterprise/adaptor/Config.java
+++ b/src/com/google/enterprise/adaptor/Config.java
@@ -599,7 +599,7 @@ public class Config {
   /**
    * Returns a map configuration the content transformer.
    */
-  List<Map<String, String>> getContentTransformerPipelineSpec() {
+  List<Map<String, String>> getContentTransformFactorySpec() {
     return getListOfConfigs("content.transformer.pipeline");
   }
 
--- a/src/com/google/enterprise/adaptor/Config.java
+++ b/src/com/google/enterprise/adaptor/Config.java
@@ -175,7 +175,7 @@ import java.util.logging.Logger;
  * <tr><td> </td><td>transform.pipeline </td><td> sequence of
  *     transformation steps.  Defaults to no-pipeline
  * <tr><td> </td><td>content.transform.pipeline </td><td>
- *     content transformer pipeline. Defaults to no-pipeline</td></tr>
+ *     content transform pipeline. Defaults to no-pipeline</td></tr>
  * <tr><td> </td><td>saml.idpExpirationMillis </td><td> Expiration time
  *     sent in SAML Authentication response. Defaults to 30,000 milliseconds.
  * </table>
@@ -600,7 +600,7 @@ public class Config {
    * Returns a map configuration the content transformer.
    */
   List<Map<String, String>> getContentTransformFactorySpec() {
-    return getListOfConfigs("content.transformer.pipeline");
+    return getListOfConfigs("content.transform.pipeline");
   }
 
   /**
--- a/src/com/google/enterprise/adaptor/Config.java
+++ b/src/com/google/enterprise/adaptor/Config.java
@@ -597,7 +597,7 @@ public class Config {
   }
 
   /**
-   * Returns a map configuration the content transform pipeline.
+   * Returns a map configuration of the content transform pipeline.
    */
   List<Map<String, String>> getContentTransformFactorySpec() {
     return getListOfConfigs("content.transform.pipeline");
--- a/src/com/google/enterprise/adaptor/Config.java
+++ b/src/com/google/enterprise/adaptor/Config.java
@@ -174,8 +174,8 @@ import java.util.logging.Logger;
  *     and modify principals as described. Defaults no modifications
  * <tr><td> </td><td>transform.pipeline </td><td> sequence of
  *     transformation steps.  Defaults to no-pipeline
- * <tr><td> </td><td>content.transformer.pipeline </td><td> content transformer pipeline.
- *     Defaults to no-pipeline</td></tr>
+ * <tr><td> </td><td>content.transformer.pipeline </td><td>
+ *     content transformer pipeline. Defaults to no-pipeline</td></tr>
  * <tr><td> </td><td>saml.idpExpirationMillis </td><td> Expiration time
  *     sent in SAML Authentication response. Defaults to 30,000 milliseconds.
  * </table>
--- a/src/com/google/enterprise/adaptor/Config.java
+++ b/src/com/google/enterprise/adaptor/Config.java
@@ -174,6 +174,8 @@ import java.util.logging.Logger;
  *     and modify principals as described. Defaults no modifications
  * <tr><td> </td><td>transform.pipeline </td><td> sequence of
  *     transformation steps.  Defaults to no-pipeline
+ * <tr><td> </td><td>content.transformer.pipeline </td><td> content transformer pipeline.
+ *     Defaults to no-pipeline</td></tr>
  * <tr><td> </td><td>saml.idpExpirationMillis </td><td> Expiration time
  *     sent in SAML Authentication response. Defaults to 30,000 milliseconds.
  * </table>
@@ -286,6 +288,7 @@ public class Config {
     addKey("adaptor.docContentTimeoutSecs", "180");
     addKey("adaptor.docHeaderTimeoutSecs", "30");
     addKey("transform.pipeline", "");
+    addKey("content.transformer.pipeline", "");
     addKey("journal.reducedMem", "true");
     addKey("gsa.acceptsDocControlsHeader", "GENERATE", new ValueComputer() {
           public String compute(String rawValue) {
@@ -591,6 +594,13 @@ public class Config {
    */
   List<Map<String, String>> getTransformPipelineSpec() {
     return getListOfConfigs("transform.pipeline");
+  }
+
+  /**
+   * Returns a map configuration the content transformer.
+   */
+  List<Map<String, String>> getContentTransformerPipelineSpec() {
+    return getListOfConfigs("content.transformer.pipeline");
   }
 
   /**
--- a/src/com/google/enterprise/adaptor/ContentTransformFactory.java
+++ b/src/com/google/enterprise/adaptor/ContentTransformFactory.java
@@ -41,7 +41,7 @@ public class ContentTransformFactory {
     if (contentTransformers.size() <= 0) {
       return original;
     }
-    DocumentContentTransformer contentTransformer = null;
+    DocumentContentTransform contentTransform = null;
     for (int i = contentTransformers.size(); i >= 0; i--) {
       final Map<String, String> tConfig = contentTransformers.get(i);
       final String className = tConfig.get("class");
@@ -53,25 +53,25 @@ public class ContentTransformFactory {
       }
       try {
         //noinspection unchecked
-        final Class<DocumentContentTransformer> clazz =
-            (Class<DocumentContentTransformer>) Class.forName(className);
-        final Constructor<DocumentContentTransformer> constructor =
+        final Class<DocumentContentTransform> clazz =
+            (Class<DocumentContentTransform>) Class.forName(className);
+        final Constructor<DocumentContentTransform> constructor =
             clazz.getConstructor(Map.class,
                 OutputStream.class, String.class, Metadata.class);
-        if (null == contentTransformer) {
-          contentTransformer =
+        if (null == contentTransform) {
+          contentTransform =
               constructor.newInstance(tConfig, original, contentType,
                   metadata);
         } else {
-          contentTransformer =
-              constructor.newInstance(tConfig, contentTransformer,
+          contentTransform =
+              constructor.newInstance(tConfig, contentTransform,
                   contentType, metadata);
         }
       } catch (Exception e) {
         throw new RuntimeException(
-            "Cannot use document content transformer of type " + className);
+            "Cannot use document content transform of type " + className);
       }
     }
-    return contentTransformer;
+    return contentTransform;
   }
 }
--- a/src/com/google/enterprise/adaptor/ContentTransformFactory.java
+++ b/src/com/google/enterprise/adaptor/ContentTransformFactory.java
@@ -42,7 +42,7 @@ public class ContentTransformFactory {
       final String className = tConfig.get("class");
       if (Strings.isNullOrEmpty(className)) {
         throw new InvalidConfigurationException(
-            "Document Content Transform class is missing " + className);
+            "Document Content Transform class is missing: " + className);
       }
       try {
         //noinspection unchecked
@@ -53,7 +53,7 @@ public class ContentTransformFactory {
         this.transforms.put(constructor, tConfig);
       } catch (Exception e) {
         throw new RuntimeException(
-            "Cannot get document content transform of type " + className, e);
+            "Cannot get document content transform of type: " + className, e);
       }
     }
   }
@@ -88,7 +88,7 @@ public class ContentTransformFactory {
         last = transform;
       } catch (Exception e) {
         throw new RuntimeException(
-            "Cannot instantiate document content transform "
+            "Cannot instantiate document content transform: "
                 + t.getKey().getName(), e);
       }
     }
--- a/src/com/google/enterprise/adaptor/ContentTransformFactory.java
+++ b/src/com/google/enterprise/adaptor/ContentTransformFactory.java
@@ -42,7 +42,8 @@ public class ContentTransformFactory {
       return original;
     }
     DocumentContentTransformer contentTransformer = null;
-    for (final Map<String, String> tConfig : contentTransformers) {
+    for (int i = contentTransformers.size(); i >= 0; i--) {
+      final Map<String, String> tConfig = contentTransformers.get(i);
       final String className = tConfig.get("class");
       if (Strings.isNullOrEmpty(className)) {
         LOG.log(Level.SEVERE,
--- a/src/com/google/enterprise/adaptor/ContentTransformFactory.java
+++ b/src/com/google/enterprise/adaptor/ContentTransformFactory.java
@@ -4,12 +4,9 @@ import com.google.common.base.Strings;
 
 import java.io.OutputStream;
 import java.lang.reflect.Constructor;
-import java.lang.reflect.InvocationTargetException;
-import java.util.LinkedList;
+import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.logging.Level;
-import java.util.logging.Logger;
 
 /**
  * The content transform factory holds all document content transformers
@@ -19,14 +16,16 @@ import java.util.logging.Logger;
  */
 public class ContentTransformFactory {
 
-  private List<DocumentContentTransform> transforms;
+  private Map<Constructor<DocumentContentTransform>, Map<String, String>>
+      transforms;
 
   public ContentTransformFactory(
       final List<Map<String, String>> transforms) {
     if (transforms.size() <= 0) {
       return;
     }
-    this.transforms = new LinkedList<DocumentContentTransform>();
+    this.transforms = new LinkedHashMap<Constructor<DocumentContentTransform>,
+        Map<String, String>>();
     for (int i = (transforms.size() - 1); i >= 0; i--) {
       final Map<String, String> tConfig = transforms.get(i);
       final String className = tConfig.get("class");
@@ -40,10 +39,10 @@ public class ContentTransformFactory {
             (Class<DocumentContentTransform>) Class.forName(className);
         final Constructor<DocumentContentTransform> constructor =
             clazz.getConstructor(Map.class);
-        this.transforms.add(constructor.newInstance(tConfig));
+        this.transforms.put(constructor, tConfig);
       } catch (Exception e) {
         throw new RuntimeException(
-            "Cannot use document content transform of type " + className, e);
+            "Cannot get document content transform of type " + className, e);
       }
     }
   }
@@ -63,15 +62,24 @@ public class ContentTransformFactory {
       return original;
     }
     DocumentContentTransform last = null;
-    for (final DocumentContentTransform transform : transforms) {
-      if (null == last) {
-        transform.setOriginalStream(original);
-      } else {
-        transform.setOriginalStream(last);
+    for (Map.Entry<Constructor<DocumentContentTransform>, Map<String, String>>
+        t : transforms.entrySet()) {
+      try {
+        final DocumentContentTransform transform =
+            t.getKey().newInstance(t.getValue());
+        if (null == last) {
+          transform.setOriginalStream(original);
+        } else {
+          transform.setOriginalStream(last);
+        }
+        transform.setContentType(contentType);
+        transform.setMetadata(metadata);
+        last = transform;
+      } catch (Exception e) {
+        throw new RuntimeException(
+            "Cannot instantiate document content transform "
+                + t.getKey().getName(), e);
       }
-      transform.setContentType(contentType);
-      transform.setMetadata(metadata);
-      last = transform;
     }
     return last;
   }
--- a/src/com/google/enterprise/adaptor/ContentTransformFactory.java
+++ b/src/com/google/enterprise/adaptor/ContentTransformFactory.java
@@ -68,8 +68,6 @@ public class ContentTransformFactory {
                   contentType, metadata);
         }
       } catch (Exception e) {
-        LOG.log(Level.SEVERE,
-            "Cannot use document content transformer of type {0}", className);
         throw new RuntimeException(
             "Cannot use document content transformer of type " + className);
       }
--- a/src/com/google/enterprise/adaptor/ContentTransformFactory.java
+++ b/src/com/google/enterprise/adaptor/ContentTransformFactory.java
@@ -50,7 +50,8 @@ public class ContentTransformFactory {
         final Class<DocumentContentTransform> clazz =
             (Class<DocumentContentTransform>) Class.forName(className);
         final Constructor<DocumentContentTransform> constructor =
-            clazz.getConstructor(Map.class);
+            clazz.getConstructor(Map.class, Metadata.class,
+                String.class, OutputStream.class);
         this.transforms.put(constructor, new TreeMap<String, String>(config));
       } catch (Exception e) {
         throw new InvalidConfigurationException(
@@ -77,16 +78,13 @@ public class ContentTransformFactory {
     for (Map.Entry<Constructor<DocumentContentTransform>, Map<String, String>>
         t : transforms.entrySet()) {
       try {
-        final DocumentContentTransform transform =
-            t.getKey().newInstance(t.getValue());
         if (null == last) {
-          transform.setOriginalStream(original);
+          last = t.getKey()
+              .newInstance(t.getValue(), metadata, contentType, original);
         } else {
-          transform.setOriginalStream(last);
+          last = t.getKey()
+              .newInstance(t.getValue(), metadata, contentType, last);
         }
-        transform.setContentType(contentType);
-        transform.setMetadata(metadata);
-        last = transform;
       } catch (Exception e) {
         throw new RuntimeException(
             "Cannot instantiate document content transform: "
--- a/src/com/google/enterprise/adaptor/ContentTransformFactory.java
+++ b/src/com/google/enterprise/adaptor/ContentTransformFactory.java
@@ -9,7 +9,7 @@ import java.util.List;
 import java.util.Map;
 
 /**
- * The content transform factory holds all document content transformers
+ * The content transform factory holds all document content transforms
  * and puts them in series connection.
  *
  * @author Dominik Weidenfeld (dominik.weidenfeld@twt.de)
--- a/src/com/google/enterprise/adaptor/ContentTransformFactory.java
+++ b/src/com/google/enterprise/adaptor/ContentTransformFactory.java
@@ -7,6 +7,7 @@ import java.lang.reflect.Constructor;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.TreeMap;
 
 /**
  * The content transform factory holds all document content transforms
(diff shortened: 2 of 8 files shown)


PapaJoltOmega:
I've added a few comments to the CL.

Thank you,
PJ

dweidenfeld:
I have made the code changes. Please check it again. The code has been cleaned up and is now simpler that before.

src/com/google/enterprise/adaptor/Config.java
```diff
@@ -174,6 +174,8 @@
  *     and modify principals as described. Defaults no modifications
  * <tr><td> </td><td>transform.pipeline </td><td> sequence of
  *     transformation steps.  Defaults to no-pipeline
+ * <tr><td> </td><td>content.transformer.pipeline </td><td> content transformer pipeline.
```
PapaJoltOmega: over 80 characters in length

src/com/google/enterprise/adaptor/DocumentContentTransformer.java
```diff
@@ -0,0 +1,58 @@
...
+ */
+public class DocumentContentTransformer extends OutputStream {
+
+  protected final Map<String, String> config;
+  protected final Metadata metadata;
+  protected final String contentType;
+  private OutputStream originalStream;
+
+  public DocumentContentTransformer(final Map<String, String> config,
+                                    final OutputStream originalStream, final String contentType,
```
pjogooglecom: strange to have one var per line and then two on this one.

src/com/google/enterprise/adaptor/DocumentContentTransformer.java
```diff
@@ -0,0 +1,58 @@
...
+                                    final Metadata metadata) {
+    // config and metadata cannot be null
```
PapaJoltOmega: add if null checks on them and throw NPE

src/com/google/enterprise/adaptor/DocumentContentTransformer.java
```diff
@@ -0,0 +1,58 @@
...
+    // contentType possibly be null, but it is allowed
```
PapaJoltOmega: skip the inlined comment.  if it is not checked then it will be allowed.
possibly add comment to ctor's javadoc.

src/com/google/enterprise/adaptor/DocumentContentTransformer.java
```diff
@@ -0,0 +1,58 @@
...
+    if (null == originalStream) {
+      throw new NullPointerException();
+    }
+    this.config = Collections.unmodifiableMap(config);
```
PapaJoltOmega: copy before applying unmodifiable, so that calleer cannot change config on us.

src/com/google/enterprise/adaptor/DocumentContentTransformerPipeline.java
```diff
@@ -0,0 +1,70 @@
...
+ * and puts them in series connection.
+ *
+ * @author Dominik Weidenfeld (dominik.weidenfeld@twt.de)
+ */
+public class DocumentContentTransformerPipeline {
+
+  private static final Logger LOG =
+      Logger.getLogger(DocumentContentTransformerPipeline.class.getName());
+
+  private List<Map<String, String>> contentTransformers;
```
pjogooglecom: why are we not storing List<DocumentContentTransform>?

src/com/google/enterprise/adaptor/DocumentContentTransformerPipeline.java
```diff
@@ -0,0 +1,70 @@
...
+                                           final String contentType,
+                                           final Metadata metadata) {
+    if (contentTransformers.size() <= 0) {
+      return original;
+    }
+    DocumentContentTransformer contentTransformer = null;
+    for (final Map<String, String> tConfig : contentTransformers) {
+      final String className = tConfig.get("class");
+      if (Strings.isNullOrEmpty(className)) {
+        LOG.log(Level.WARNING, "Document Content Transformer class is missing");
```
PapaJoltOmega: why not consider this fatal and throw RuntimeException?

src/com/google/enterprise/adaptor/DocumentContentTransformerPipeline.java
```diff
@@ -0,0 +1,70 @@
...
+            (Class<DocumentContentTransformer>) Class.forName(className);
+        final Constructor<DocumentContentTransformer> constructor =
+            clazz.getConstructor(Map.class, OutputStream.class, String.class, Metadata.class);
+        if (null == contentTransformer) {
+          contentTransformer =
+              constructor.newInstance(tConfig, original, contentType, metadata);
+        } else {
+          contentTransformer =
+              constructor.newInstance(tConfig, contentTransformer, contentType, metadata);
+        }
```
PapaJoltOmega: If someone has a config with multiple content-transforms the last specified content transform is the first one to get the bytes?  Is my understanding correct?

src/com/google/enterprise/adaptor/DocumentContentTransformerPipeline.java
```diff
@@ -0,0 +1,70 @@
...
+      } catch (Exception e) {
```
PapaJoltOmega: What exceptions show up here? Is it just the generic reflection stuff?  Why not wrap in RuntimeException and rethrow?

src/com/google/enterprise/adaptor/DocumentContentTransformerPipeline.java
```diff
@@ -0,0 +1,70 @@
...
+
+  /**
+   * Creates a new transformer pipeline.
+   *
+   * @param original    original content stream
+   * @param contentType content type
+   * @param metadata    metadata
+   * @return stream pipeline or original output stream
+   */
+  public final OutputStream createPipeline(final OutputStream original,
```
PapaJoltOmega: I see what you mean about there not needing to be call to transform().  It is because the OuputStream is wrapped and the pipeline invocation is implicit in the layering of OutputStream rather than needing an explicit call.  

Then perhaps this class is not necessary or perhaps it is a bit misnamed at least.  That is, consider that if the instance of this class is a "DocumentContentTransformerPipeline" then it should not have a method to "createPipeline" as it is already is a pipeline by virtue of its name.

Let me think of some better naming by tmrw hopefully.

src/com/google/enterprise/adaptor/DocumentContentTransformerPipeline.java
```diff
@@ -0,0 +1,70 @@
...
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+/**
+ * The document content transformer pipeline holds all document content transformers
+ * and puts them in series connection.
+ *
+ * @author Dominik Weidenfeld (dominik.weidenfeld@twt.de)
+ */
+public class DocumentContentTransformerPipeline {
```
PapaJoltOmega: maybe?
s/DocumentContentTransformerPipeline/ContentTransformFactory/

PapaJoltOmega:
I'm thinking that it will be common to use document-content-transform classes to call out to executables in the system. 

In thise case someone needs to save the entire byte content, say of a binary document format, and then they need to pass it to a binary executable on their system.  

I am thinking that this scenario could be handled by:
(1) a document-content-transform buffering all the write() calls into a byte-array-type-object like ByteArrayOutputStream
(2) Then on close() the document-content-transform would pass the entire buffer to the binary executable
(3) Then the document-content-transform would get output of the binary executable
(4) Then it would pass the entire byte content from exeutable to a write() on the OutputStream that it itself was wrapping
(5) Then it would call close on the OutputStream that it itself was wrapping

Is that what you see also?

Thank you

PapaJoltOmega:
Thank you Dominik,

I've added some comments.

I've also added some thoughts to see if I understand the configuration (the
sequnce of application of the content transforms).

I've also added some thought to see if I understand how someone would use
an external binary executable to change content.

Thank you,
PJ

src/com/google/enterprise/adaptor/Config.java
dweidenfeld: fixed

src/com/google/enterprise/adaptor/DocumentContentTransformer.java
dweidenfeld: there are already all necessary null checks. The other ones are not mandatory (see new javadoc comments)

src/com/google/enterprise/adaptor/DocumentContentTransformer.java
dweidenfeld: done

src/com/google/enterprise/adaptor/DocumentContentTransformer.java
dweidenfeld: done

src/com/google/enterprise/adaptor/DocumentContentTransformerPipeline.java
dweidenfeld: done, but now the code is not as fault tolerant ;) (I think that is what you supposed)

src/com/google/enterprise/adaptor/DocumentContentTransformerPipeline.java
dweidenfeld: Yes, you're right. I changed to order like in this example:

You have 3 Transformers configured (T1, T2, T3).
Now the object will be created like T1(T2(T3(originalStream))).
So if the adaptor pushes content to the stream, the first instance is T1, which calls T2, which calls T3, which calls the original stream.

src/com/google/enterprise/adaptor/DocumentContentTransformerPipeline.java
dweidenfeld: done

src/com/google/enterprise/adaptor/DocumentContentTransformerPipeline.java
dweidenfeld: Renaming done ;)

src/com/google/enterprise/adaptor/DocumentContentTransformerPipeline.java
dweidenfeld: see comment above

dweidenfeld:
> I'm thinking that it will be common to use document-content-transform classes to call out to executables in the system.
> 
> In thise case someone needs to save the entire byte content, say of a binary document format, and then they need to pass it to a binary executable on their system.
> 
> I am thinking that this scenario could be handled by:
> (1) a document-content-transform buffering all the write() calls into a byte-array-type-object like ByteArrayOutputStream
> (2) Then on close() the document-content-transform would pass the entire buffer to the binary executable
> (3) Then the document-content-transform would get output of the binary executable
> (4) Then it would pass the entire byte content from exeutable to a write() on the OutputStream that it itself was wrapping
> (5) Then it would call close on the OutputStream that it itself was wrapping
> 
> Is that what you see also?
> 
> Thank you

Yeah. you're right, this is a possible behaviour. It could possibly affect the adaptor performance and memory usage if someone writes a content transformer that holds all the binary information till the stream gets closed. I think it is responsibility of the programmer, that writes such a transformer, to take care of performance and memory usage.

In my case (this is why I added this function), I will get the content on every write and pass it over a JMS connection to a second program, that creates things like thumbnails for image binaries.

src/com/google/enterprise/adaptor/Config.java
```diff
@@ -286,6 +288,7 @@ public String compute(String rawValue) {
     addKey("adaptor.docContentTimeoutSecs", "180");
     addKey("adaptor.docHeaderTimeoutSecs", "30");
     addKey("transform.pipeline", "");
```
PapaJoltOmega, line 290: I'm thinking that after this CL is submitted I'll rename "transform.pipeline" to "metadata.transform.pipeline" and also rename classes in the source code base.  For backwards compatability I'll have the code continue to accept "transform.pipeline" for "metadata.transform.pipeline", but the class names will be redone.

What do you think?

src/com/google/enterprise/adaptor/Config.java
```diff
@@ -286,6 +288,7 @@ public String compute(String rawValue) {
...
+    addKey("content.transformer.pipeline", "");
```
PapaJoltOmega: s/transformer/transform/

src/com/google/enterprise/adaptor/ContentTransformFactory.java
```diff
@@ -0,0 +1,78 @@
...
+          contentTransformer =
+              constructor.newInstance(tConfig, contentTransformer,
+                  contentType, metadata);
+        }
+      } catch (Exception e) {
+        LOG.log(Level.SEVERE,
+            "Cannot use document content transformer of type {0}", className);
+        throw new RuntimeException(
+            "Cannot use document content transformer of type " + className);
+      }
```
PapaJoltOmega, line 92: remove the log; leave the throw please

src/com/google/enterprise/adaptor/ContentTransformFactory.java
```diff
@@ -0,0 +1,78 @@
...
+                OutputStream.class, String.class, Metadata.class);
+        if (null == contentTransformer) {
+          contentTransformer =
+              constructor.newInstance(tConfig, original, contentType,
+                  metadata);
+        } else {
+          contentTransformer =
+              constructor.newInstance(tConfig, contentTransformer,
+                  contentType, metadata);
+        }
```
PapaJoltOmega, line 87: I think if someone specifies transforms=T1,T2,T3 then T1 should get the bytes first, then T2 and then T3.

I think the two pipelines (the metadata one and the content one) should go in the same order and the metadata one goes from left to right already.

src/com/google/enterprise/adaptor/DocumentContentTransformer.java
```diff
@@ -0,0 +1,68 @@
...
+   *
+   * @param config         the configuration for this instance
+   *                       (never null, but maybe empty)
+   * @param originalStream the original content stream
+   * @param contentType    the content type of the stream
+   *                       (could possibly be null or empty)
+   * @param metadata       the metadata already collected for this document
+   *                       (never null, but maybe empty)
+   */
+  public DocumentContentTransformer(final Map<String, String> config,
```
PapaJoltOmega: s/Transformer/Transform/

I don't have anything against Transformer as a word, but there are already AclTransform and DocumentTransform (which will become MetadataTransform). So I would like to keep the word the same, please.

PapaJoltOmega:
> It could possibly affect the adaptor performance 
> and memory usage if someone writes a content 
> transformer that holds all the binary information till 
> the stream gets closed.  I think it is responsibility 
> of the programmer, that writes such a transformer, 
> to take care of performance and memory usage.

I agree.  I am also not concerned about the memory.
Sorry for not being clear, I was trying to confirm that
the use case of using external binary would be 
possible and checking that what I was envisioning
would work.

src/com/google/enterprise/adaptor/ContentTransformFactory.java
```diff
@@ -0,0 +1,78 @@
...
+    if (contentTransformers.size() <= 0) {
+      return original;
+    }
+    DocumentContentTransformer contentTransformer = null;
+    for (final Map<String, String> tConfig : contentTransformers) {
+      final String className = tConfig.get("class");
+      if (Strings.isNullOrEmpty(className)) {
+        LOG.log(Level.SEVERE,
+            "Document Content Transformer class is missing {0}", className);
+        throw new RuntimeException(
```
PapaJoltOmega: just throw; don't log.

test/com/google/enterprise/adaptor/DocumentHandlerTest.java
```diff
@@ -1693,6 +1729,25 @@ public void testPercentEncoding() {
                  + "%2A%28%29%5B%5D%7B%7D%C3%AB%01", encoded);
   }
 
+  private static class SampleDocumentContentTransformer
```
PapaJoltOmega: consider placing this inner class closer to testContentTransformer() method

PapaJoltOmega:
I believe we are close to submitting this CL.

Thank you,
PJ

src/com/google/enterprise/adaptor/Config.java
dweidenfeld, line 290: Jeah. I think this is the best way to go.

src/com/google/enterprise/adaptor/Config.java
dweidenfeld: Oh, seems to missed that. Fixed ;)

src/com/google/enterprise/adaptor/ContentTransformFactory.java
dweidenfeld, line 92: done

src/com/google/enterprise/adaptor/ContentTransformFactory.java
dweidenfeld, line 87: I changed the working order (see for loop call). But I missed to push the changes yesterday ;)
Now the for loop is in reverse order, so that it handles it like T1,T2,T3 and T1 is the outermost wrapper, so that it will get the content first.

src/com/google/enterprise/adaptor/DocumentContentTransformer.java
dweidenfeld: No problem. I only missed the renaming. Done.

src/com/google/enterprise/adaptor/ContentTransformFactory.java
dweidenfeld: done

test/com/google/enterprise/adaptor/DocumentHandlerTest.java
dweidenfeld: I added the testing class next to all other inner classes, so I thought it would be correct. Now I moved it next to the test function.

src/com/google/enterprise/adaptor/ContentTransformFactory.java
```diff
@@ -0,0 +1,75 @@
...
+ * and puts them in series connection.
+ *
+ * @author Dominik Weidenfeld (dominik.weidenfeld@twt.de)
+ */
+public class ContentTransformFactory {
+
+  private static final Logger LOG =
+      Logger.getLogger(ContentTransformFactory.class.getName());
+
+  private List<Map<String, String>> contentTransformers;
```
PapaJoltOmega: Let's store List of Constructor of DocumentContentTransform

Let's have the ContentTransformFactory constructor go thru the config and call Class.forName().

This way we will catch many incorrect configratiuons at creation of this class' instance rather then when they are attempted to be used inside createPipeline call.

We will still need to keep config because that is first parameter to DocumentContentTransform ctor.  But let's keep copy of the passed in config paramter, rather than just reference to it.

src/com/google/enterprise/adaptor/ContentTransformFactory.java
```diff
@@ -0,0 +1,75 @@
...
+              constructor.newInstance(tConfig, original, contentType,
+                  metadata);
+        } else {
+          contentTransform =
+              constructor.newInstance(tConfig, contentTransform,
+                  contentType, metadata);
+        }
+      } catch (Exception e) {
+        throw new RuntimeException(
+            "Cannot use document content transform of type " + className);
```
PapaJoltOmega: chain the Exception e to RuntimeException

src/com/google/enterprise/adaptor/DocumentContentTransform.java
```diff
@@ -0,0 +1,68 @@
...
+   *                       (never null, but maybe empty)
+   * @param originalStream the original content stream
+   * @param contentType    the content type of the stream
+   *                       (could possibly be null or empty)
+   * @param metadata       the metadata already collected for this document
+   *                       (never null, but maybe empty)
+   */
+  public DocumentContentTransform(final Map<String, String> config,
+                                  final OutputStream originalStream, final String contentType,
+                                  final Metadata metadata) {
```
PapaJoltOmega: Metadata instances are mutable.

I'm thinking whether programmers of DocumentContentTransform classes will expect to be able to change it.

At the point where a DocumentContentTransform is instantiated the HTTP headers have already been sent out, so any changes won't affect the metadata GSA receives.  However it could be used as a communication channel between DocumentContentTransform instances in a pipeline.  I wonder if that is a good thing, or if that is something that should be made more explicit.

src/com/google/enterprise/adaptor/DocumentContentTransform.java
```diff
@@ -0,0 +1,68 @@
...
+  @Override
+  public void flush() throws IOException {
+    originalStream.flush();
+  }
+
+  @Override
+  public void close() throws IOException {
+    originalStream.close();
+  }
+}
```
PapaJoltOmega, line 76: consider adding a toString method.  Also consider changing the HashMap to a TreeMap, so then toString's showing of config will be alphabetical.

PapaJoltOmega:
Thank you for your patiance.  Further comments.

src/com/google/enterprise/adaptor/ContentTransformFactory.java
dweidenfeld: I changed the implementation, so that all transformers are instantiated in the constructor.
The other "non-optional" parameters are now passed by setters, that guarantees their correctness.

src/com/google/enterprise/adaptor/ContentTransformFactory.java
dweidenfeld: done

src/com/google/enterprise/adaptor/DocumentContentTransform.java
dweidenfeld: What would you recommend.
I think that this communication channel could be useful.

But maybe I could introduce some kind of getMetadataValue(String key); and getMetadataValues(String key); methods and make the metadata object private scope, so that the transformer can only access the values and not the actual metadata object.

What do you think?

src/com/google/enterprise/adaptor/DocumentContentTransform.java
dweidenfeld, line 76: done

src/com/google/enterprise/adaptor/ContentTransformFactory.java
```diff
@@ -0,0 +1,78 @@
...
+        throw new RuntimeException(
+            "Document Content Transform class is missing " + className);
+      }
+      try {
+        //noinspection unchecked
+        final Class<DocumentContentTransform> clazz =
+            (Class<DocumentContentTransform>) Class.forName(className);
+        final Constructor<DocumentContentTransform> constructor =
+            clazz.getConstructor(Map.class);
+        this.transforms.add(constructor.newInstance(tConfig));
```
PapaJoltOmega: This change is great, but newInstance should be in createPipeline.

The various threads calling getDocContent cannot be using the
same instances of DocumentContentTransform.

Thank you

src/com/google/enterprise/adaptor/Config.java
```diff
@@ -594,6 +597,13 @@ long getAdaptorDocContentTimeoutMillis() {
   }
 
   /**
+   * Returns a map configuration the content transformer.
```
PapaJoltOmega: s/the content transformer/of the content transform pipeline/

src/com/google/enterprise/adaptor/ContentTransformFactory.java
```diff
@@ -0,0 +1,78 @@
...
+      final List<Map<String, String>> transforms) {
+    if (transforms.size() <= 0) {
+      return;
+    }
+    this.transforms = new LinkedList<DocumentContentTransform>();
+    for (int i = (transforms.size() - 1); i >= 0; i--) {
+      final Map<String, String> tConfig = transforms.get(i);
+      final String className = tConfig.get("class");
+      if (Strings.isNullOrEmpty(className)) {
+        throw new RuntimeException(
```
PapaJoltOmega: throw InvalidConfigurationException instead please

https://github.com/googlegsa/library/blob/master/src/com/google/enterprise/adaptor/InvalidConfigurationException.java

Also add javadoc for this ctor that mentions excpetion is thrown if config is bad

src/com/google/enterprise/adaptor/ContentTransformFactory.java
```diff
@@ -0,0 +1,78 @@
...
+            "Document Content Transform class is missing " + className);
```
PapaJoltOmega: s/missing/missing:/

test/com/google/enterprise/adaptor/DocumentHandlerTest.java
```diff
@@ -402,6 +406,56 @@ public void getDocContent(Request request, Response response)
...
+  private static class SampleDocumentContentTransform
+      extends DocumentContentTransform {
+
+    public SampleDocumentContentTransform(Map<String, String> config) {
+      super(config);
+    }
+
+    @Override
+    public void write(final byte[] b) throws IOException {
+      if (contentType.equals("image/jpeg")) {
```
PapaJoltOmega, line 453: can the content transform change the contet-type string?

That is say we put in a PDF and content transform makes it a text file.  Can we have the content transform put back the content-type string in DocumentHandler?

Let's not do that in this CL, but let's think about whether it is doable, and leave a TODO. One thing that may be a challange is that we need to know the resulting content type before we send a single byte of the response body (because the content-type ahs to be in HTTP header).

I think it will be fine, i think we will be able to have the content transform do something like:

String getContentTypeOutputForContentTypeInput(String ctIn) {
  if ("text/PDF".equalsIgnoreCase(ctIn)) {
    return "text/plain"
  }
  return ctIn;
}

src/com/google/enterprise/adaptor/DocumentContentTransform.java
PapaJoltOmega, line 76: Great.  Thank you.

I've left four comments for you.
1 is important, about newInstance call needing to be in createPipeline

1 is thinking about the future to let transform change content-type variable

2 are requests for rewording comments.

Thank you very much,
PJ

PapaJoltOmega, line 76: I've got a CL that will be changing DocumentHandler
https://codereview.appspot.com/261770043/.

I'll wait until we've submitted your content transform
CL before submitting the one I've got going.  That way
I'll get to resolve any emergent merge conflict.

Thank you,
PJ

src/com/google/enterprise/adaptor/ContentTransformFactory.java
dweidenfeld: Oh.. great you noticed it. Fixed in new commit.

src/com/google/enterprise/adaptor/Config.java
dweidenfeld: done

src/com/google/enterprise/adaptor/ContentTransformFactory.java
dweidenfeld: done

src/com/google/enterprise/adaptor/ContentTransformFactory.java
dweidenfeld: done

test/com/google/enterprise/adaptor/DocumentHandlerTest.java
dweidenfeld, line 453: I have added the TODO, but we need to take care that the contentTransformer.getFinalContentType() (or similar) method is called before startSending().
Maybe you want to watch the binary content first before changing the content-type... this would not be possible...

``` java
    if ("HEAD".equals(ex.getRequestMethod())) {
        // Unfortunately, we won't be able to report any errors after this
        // point. We don't delay sending the headers, however, because of the
        // watchdog.
        state = State.HEAD;
        startSending(false);
        os = new SinkOutputStream();
      } else {
        state = State.SEND_BODY;
        startSending(true);
        countingOs = new CountingOutputStream(new CloseNotifyOutputStream(
            ex.getResponseBody()));
        os = countingOs;
      }
      if (null != contentTransformFactory) {
        // TODO: possibility for changing the content-type in transformer
        return contentTransformFactory
            .createPipeline(os, contentType, metadata);
      }
```

src/com/google/enterprise/adaptor/ContentTransformFactory.java
```diff
@@ -0,0 +1,97 @@
...
+      try {
+        //noinspection unchecked
+        final Class<DocumentContentTransform> clazz =
+            (Class<DocumentContentTransform>) Class.forName(className);
+        final Constructor<DocumentContentTransform> constructor =
+            clazz.getConstructor(Map.class);
+        this.transforms.put(constructor, tConfig);
+      } catch (Exception e) {
+        throw new RuntimeException(
+            "Cannot get document content transform of type: " + className, e);
```
PapaJoltOmega, line 58: Also change to ICE

src/com/google/enterprise/adaptor/ContentTransformFactory.java
```diff
@@ -0,0 +1,97 @@
...
+   * instantiation.
+   *
+   * @param transforms Configuration for each {@link DocumentContentTransform}
+   * @throws InvalidConfigurationException If the class for a {@link
+   *                                       DocumentContentTransform} is missing
+   * @throws RuntimeException              If the class does not match all
+   *                                       criteria
+   */
+  public ContentTransformFactory(
+      final List<Map<String, String>> transforms) {
```
PapaJoltOmega: although it's common for ctors to use the same name of param and inner class variable, it's atypical and a bit confusing to do so in this method.  Please change parameter.  Perhaps change transforms to configs or confugations.

src/com/google/enterprise/adaptor/ContentTransformFactory.java
```diff
@@ -0,0 +1,97 @@
...
+        throw new InvalidConfigurationException(
+            "Document Content Transform class is missing: " + className);
+      }
+      try {
+        //noinspection unchecked
+        final Class<DocumentContentTransform> clazz =
+            (Class<DocumentContentTransform>) Class.forName(className);
+        final Constructor<DocumentContentTransform> constructor =
+            clazz.getConstructor(Map.class);
+        this.transforms.put(constructor, tConfig);
```
PapaJoltOmega: make a copy of tConfig.

test/com/google/enterprise/adaptor/DocumentHandlerTest.java
PapaJoltOmega, line 453: Here is an idea (let me know what you think):

To ContentTransformFactory add method:
   String calculateResultingType(String ct);

To ContentTransformFactory ctor add logic to check whether a DocumentContentTransform has static method with signature:
public static String getResultingType(String ct) 

Have Factory calculateResultingType call each getResultingType in turn.
If a DocumentContentTransform does not have a getResultingType static method then consider that a no-op.

Do the figuring out of whether there is getResultingType in ctor of ContentTransformFactory, do the calling in calculateResultingType.

What do you think of this idea? (it does not ahve to be put into this CL).

src/com/google/enterprise/adaptor/Config.java
```diff
@@ -594,6 +597,13 @@ long getAdaptorDocContentTimeoutMillis() {
   }
 
   /**
+   * Returns a map configuration the content transform pipeline.
```
PapaJoltOmega: s/the/of the/

src/com/google/enterprise/adaptor/DocumentContentTransform.java
```diff
@@ -0,0 +1,79 @@
...
+
+  public final void setMetadata(final Metadata metadata) {
+    this.metadata = metadata;
+  }
+
+  public final void setContentType(final String contentType) {
+    this.contentType = contentType;
+  }
+
+  public final void setOriginalStream(final OutputStream originalStream) {
```
PapaJoltOmega: what's the reason not to put the above three parameters (metadata, contentType, originalStream) into the ctor of this class?

ALl the set methods are always called by the factory, so why not have them all in ctor and avoid this class being mutable?

test/com/google/enterprise/adaptor/DocumentHandlerTest.java
PapaJoltOmega, line 453: Thank you. Added further comments.

src/com/google/enterprise/adaptor/ContentTransformFactory.java
dweidenfeld, line 58: I do not really think that this is a ICE, because it happens while the class finding process, but I changed it to ICE now ;)

src/com/google/enterprise/adaptor/ContentTransformFactory.java
dweidenfeld: alright. fixed.

src/com/google/enterprise/adaptor/ContentTransformFactory.java
dweidenfeld: done

src/com/google/enterprise/adaptor/Config.java
dweidenfeld: done

src/com/google/enterprise/adaptor/DocumentContentTransform.java
dweidenfeld: It was made because of our previous change (newInstance in ctor). That time there was no other way to pass the information to this class.
Now I added everything to the ctor again, because the newInstance is not called in the ctor anymore.

PapaJoltOmega:
LGTM.

Thank you for your patiance.  Thank you for your work.

I believe there is a trivial merge needed on your part before I can pull this request as a commit.

Thank you.

PapaJoltOmega:
Never mind. 

I merged it and pushed it.

Thank you

dweidenfeld:
Thank you for the great collaboration. I really enjoyed your feedback.

PR #8, opened 2015-08-21
Pass sensitive valuevia command line

bithazard:
+Added possibility to pass a sensitive value for obfuscation or encryption via command line

### DIFF
--- a/src/com/google/enterprise/adaptor/SensitiveValueCodec.java
+++ b/src/com/google/enterprise/adaptor/SensitiveValueCodec.java
@@ -256,7 +256,8 @@ class SensitiveValueCodec implements SensitiveValueDecoder {
    * <p>This class allows adaptor administrators to get encoded sensitive value
    * from command line.
    * 
-   * Example command line to run:
+   * Example command line to run (you will be prompted to enter the sensitive
+   * value):
    * <pre>
    * java \
    * -Djavax.net.ssl.keyStore=keys.jks \
@@ -268,11 +269,27 @@ class SensitiveValueCodec implements SensitiveValueDecoder {
    * -Dserver.keyAlias=adaptor \
    * -Dserver.secure=true
    * </pre>
+   *
+   * Example command line to run (passing the sensitive value on the command
+   * line):
+   * <pre>
+   * java \
+   * -Djavax.net.ssl.keyStore=keys.jks \
+   * -Djavax.net.ssl.keyStoreType=jks \
+   * -Djavax.net.ssl.keyStorePassword=changeit \
+   * -classpath 'adaptor-20130612-withlib.jar' \
+   * com.google.enterprise.adaptor.SensitiveValueCodec \
+   * -DsecurityLevel=ENCRYPTED \
+   * -Dserver.keyAlias=adaptor \
+   * -Dserver.secure=true
+   * 'sensitive value'
+   * </pre>
    */
   public static void main(String[] args) throws IOException {
     Config config = new Config();
     config.addKey("securityLevel", SecurityLevel.PLAIN_TEXT.toString());
     config.setValue("gsa.hostname", "not-used");
+    config.setValue("server.hostname", "not-used");
     Application.autoConfig(config, args, null);
     
     SecurityLevel securityLevel =
@@ -298,15 +315,30 @@ class SensitiveValueCodec implements SensitiveValueDecoder {
       }
     }
     SensitiveValueCodec codec = new SensitiveValueCodec(keyPair);
-    
-    Console console = System.console();
-    if (console == null) {
-      log.warning("Couldn't get Console instance");
-      return;
+    String firstNonDefineParameter = getFirstNonDefineParameter(args);
+    if (firstNonDefineParameter == null) {
+      Console console = System.console();
+      if (console == null) {
+        log.warning("Couldn't get Console instance");
+        return;
+      }
+      char passwordArray[] = console.readPassword("Sensitive Value: ");
+      String password = new String(passwordArray);
+      String encodedValue = codec.encodeValue(password, securityLevel);
+      System.out.printf("Encoded value is: %s%n", encodedValue);
+    } else {
+      String encodedValue = codec.encodeValue(firstNonDefineParameter,
+              securityLevel);
+      System.out.println(encodedValue);
     }
-    char passwordArray[] = console.readPassword("Sensitive Value: ");
-    String password = new String(passwordArray);
-    String encodedValue = codec.encodeValue(password, securityLevel);
-    console.printf("Encoded value is: %s%n", encodedValue);
+  }
+
+  private static String getFirstNonDefineParameter(String[] args) {
+    for (String arg : args) {
+      if (!arg.startsWith("-D")) {
+        return arg;
+      }
+    }
+    return null;
   }
 }
--- a/src/com/google/enterprise/adaptor/SensitiveValueCodec.java
+++ b/src/com/google/enterprise/adaptor/SensitiveValueCodec.java
@@ -272,7 +272,7 @@ class SensitiveValueCodec implements SensitiveValueDecoder {
    * -Dserver.secure=true
    * </pre>
    *
-   * You can add a stdin parameter ("--stdin") causing the sensitive value to
+   * You can add a quiet parameter ("--quiet") causing the sensitive value to
    * be read from standard input instead (this can also be a pipe). With this
    * parameter the program also only outputs the encoded sensitive value
    * without any additional text.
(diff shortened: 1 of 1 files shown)


PapaJoltOmega:
I'm concerned that command lines are saved in command line history files.  If someone uses the tool with the sensitive values being passed on the command line then they will be around unencrypted in these history files.  Any thoughts on that concern?

bithazard:
I think that's a valid concern. However the previous functionality to be prompted for the sensitive value/password is still there. Now you also have the possibility to call this function from a script (from where commands are usually not saved to command line history) or a GUI application.

Another thought: I guess most users will use the dashboard for obfuscation/encryption anyway. So it would be good to have this alternative for the usage in a script or from a GUI application.

PapaJoltOmega:
(1) The previous functionality (the interactive command line mode) does not result in sensitive info being saved in history log files. So previous functioanlity doesn't have this security problem.

(2) I envision encrypting sensitive values is a rare operation. Encrypt the sensitive value ones and use the encrupted value in config file.  Why would someone do it from a script? (which suggests some regular or repeated need to encrypt the same sensitive value). Please help me understand.

Thank you,
PJ

BrettMichaelJohnson:
IIRC, the v3 installers encrypt the passwords via a script:
encryptpassword -p
—
Brett M. Johnson

bithazard:
(1) Right, but that's not what I meant. No one is forcing you to pass the sensitive value via command line. You still have the option to use the interactive command line mode as before.

(2) I would also say it's a rare operation you do once at the beginning when you setup a new adaptor. I was in fact trying to use it in an installer I am currently building. I tried for several days to somehow write the password to the stdin of the process, sort of faking the input of the password by the user. But as soon as I write anything to the stdin Java can apparently not longer get the console instance and I just get the error "Couldn't get Console instance" from the adaptor. This seems to be a larger problem in Java that several IDEs also have. The only way I could find to circumvent this was to simply pass the sensitive value on the command line.

PapaJoltOmega:
Would it be sufficient to allow someone to pass a file that has as its contents the value that is to be encrypted?

Thank you

bithazard:
Well this would certainly be a way to prevent the sensitive value from ending up in the command line history. Honestly I don't think that I would use it though. It's quite cumbersome to first save the value to a file, pass the path to that file on the command line and delete the file afterwards. 

But it would also be OK for me if you simply not merge the changes as i understand your concerns. I would than add this functionality to each of our adaptors. I just thought, when I have to add a functionality to all of our adaptors, a better place might be the adaptor library, as this is something common that might be useful for other users as well.

BrettMichaelJohnson:
I think saving it to a file is even worse, because you risk leaving
a file with the unencrypted password in the filesystem, or a 
deleted file that might get recovered. I would prefer supplying 
the unencrypted password via stdin and collecting the encrypted
result on stdout or stderr.

The library can be modified to check if System.console() is null,
then it can avoid prompting, expect the unencrypted value on
stdin, and write the encrypted value to stdout.

Redirecting stdio is trivial in command shell scripts and only 
slightly more difficult than traditional file I/O when forking a 
new process.

—
Brett M. Johnson

PapaJoltOmega:
So for putting on stdin, how about if first argument is "--quiet" we read stdin, then encode, then write to stdout.

BrettMichaelJohnson:
I’m partial to basing it on System.console() returning null,
which only happens if stdio has been redirected.
Forcing it via command line switch isn’t terrible, though.
—
Brett M. Johnson

bithazard:
I added a new commit with the discussed changes. I'm now checking if a "--quiet" parameter is present. If so, the sensitive value is read from stdin. With this you can e.g. do the following:

echo password|java -cp <adaptor.jar> com.google.enterprise.adaptor.SensitiveValueCodec -DsecurityLevel=OBFUSCATED --quiet

This will then write the encoded value to stdout. If you don't pass the "--quiet" parameter, the behavior is as it was before.

Let me know what you think.

src/com/google/enterprise/adaptor/SensitiveValueCodec.java
```diff
@@ -268,11 +271,27 @@ String getPrefix() {
...
+   * java \
+   * -Djavax.net.ssl.keyStore=keys.jks \
+   * -Djavax.net.ssl.keyStoreType=jks \
+   * -Djavax.net.ssl.keyStorePassword=changeit \
+   * -classpath 'adaptor-20130612-withlib.jar' \
+   * com.google.enterprise.adaptor.SensitiveValueCodec \
+   * -DsecurityLevel=ENCRYPTED \
+   * -Dserver.keyAlias=adaptor \
+   * -Dserver.secure=true
+   * 'sensitive value'
```
PapaJoltOmega: stale comment doesn't reflect quiet mode

src/com/google/enterprise/adaptor/SensitiveValueCodec.java
```diff
@@ -268,11 +271,27 @@ String getPrefix() {
...
+   * </pre>
    */
   public static void main(String[] args) throws IOException {
     Config config = new Config();
     config.addKey("securityLevel", SecurityLevel.PLAIN_TEXT.toString());
     config.setValue("gsa.hostname", "not-used");
+    config.setValue("server.hostname", "not-used");
```
PapaJoltOmega, line 287: is this line necessary? why add this line?

src/com/google/enterprise/adaptor/SensitiveValueCodec.java
```diff
@@ -298,15 +317,33 @@ public static void main(String[] args) throws IOException {
...
-    Console console = System.console();
-    if (console == null) {
-      log.warning("Couldn't get Console instance");
-      return;
+    boolean quietParameterPresent = isParameterPresent(args, "--quiet");
+    if (quietParameterPresent) {
+      BufferedReader reader = new BufferedReader(new InputStreamReader(
+              System.in));
+      String encodedValue = codec.encodeValue(reader.readLine(),
+              securityLevel);
```
PapaJoltOmega: make the indent on this line and BufferedReader one be 4spaces rather than 8, please

PapaJoltOmega:
LGTM.  ASking for minor touchups.  Will merge then.

Thank you

src/com/google/enterprise/adaptor/SensitiveValueCodec.java
bithazard: Done with the last commit.

src/com/google/enterprise/adaptor/SensitiveValueCodec.java
bithazard: Right, I forgot to update this. I now removed the 2nd example command line, as the sensitive value is again always prompted in some way. Instead I added an info text explaining the new parameter and pointing out the potential risk when using this parameter. 

Oh, and I renamed the parameter from "quiet" to "stdin" as this name is way more applicable in my opinion. I found the same parameter to be used in a version of the passwd command (http://linux.die.net/man/1/passwd). So I believe this is a better name. If you disagree, I can also change the name back to "quiet".

src/com/google/enterprise/adaptor/SensitiveValueCodec.java
bithazard, line 287: In the class Config in line 210 the variable hostname is initialized with null. If an UnknownHostException is thrown a few lines below, this variable remains null. That being the case an InvalidConfigurationException will be thrown in line 894 when the config is validated ("validate" method) as step of the method "autoConfig" which is called in line 288 in SensitiveValueCodec. Hence I added this as config option with a dummy value (like "gsa.hostname").

PapaJoltOmega, line 287: I checked the github username bithazard
and do not see you having signed the
Google contributerd agreement.  It's possible
that it is under a corporate acceptance or
under an email address.

Have you signed the Google CLA? Do you
have a partner or deployment person that
you can point me to that would know? Or
I can have that sent to you.

Thank you,
PJ

src/com/google/enterprise/adaptor/SensitiveValueCodec.java
```diff
@@ -298,15 +310,33 @@ public static void main(String[] args) throws IOException {
       }
     }
     SensitiveValueCodec codec = new SensitiveValueCodec(keyPair);
-    
-    Console console = System.console();
-    if (console == null) {
-      log.warning("Couldn't get Console instance");
-      return;
+    boolean stdinParameterPresent = isParameterPresent(args, "--stdin");
```
PapaJoltOmega: The console typically is stdin.

So the presence of this flag does not uniquely denote stdin and further the absence of it does not communicate the absence of stdin.

I would prefer "--noconsole" (with --console being OK and same as no argument) or "--quiet".  I think the --stdin is a bit misleading because of the above reasons.

Thank you

bithazard: OK, I changed the parameter back to "--quiet".

src/com/google/enterprise/adaptor/SensitiveValueCodec.java
bithazard, line 287: The Google CLA should be under a corporate acceptance of TWT Business Solutions.

PapaJoltOmega, line 287: Thank you.
I found the Google corporate CLA information is accepted.

PR #15, opened 2016-11-10
Refining CidrAddress.java: add support for network / broadcast / netmask

szab100:
Adding support for network / broadcast / netmask and proper isInRange(..) resolution: answer yes for only valid IP addresses in subnet and NO for network or broadcast addresses.


szab100:
I signed it!

PR #16, opened 2016-12-14
Better Eclipse support

szab100:


szab100:
I re-tested this today and rebased with master changes.
Everything looks good in Eclipse:
  - New project -> Java Project from existing Ant file imports the project with correct classpath, src and test folders.
  - All 1062 tests pass: right click on test folder / Run As -> JUnit Test

wiarlawd:
So the purpose of build.eclipse.xml is to create a new Eclipse project. Is that better than have a .project and maybe .classpath file? We did have troubles with those in the past, and eventually deleted them, but just wanted to make sure we'd considered the option.

Could you add in the eventual commit message also a description of using resources instead of local files for the tests. Are these cases where we could be using JUnit 4 temp files instead?

This CL has a lot of 80 column errors, modified whitespace (no-no), and updated import rules (should just be left alone unless we're going to modify all the files in a separate CL, IMHO).

wiarlawd:
FYI, OOO for 10 days.

szab100:
Ah okay.. from tomorrow? I'll fix the columns / whitespace issues by the
time you are back..
BTW, what's wrong with those whitespace deletions? Our Eclipse is set up to
auto-fix those, but should just do it at lines we touch..
(the >80 column issue is probably done by that too, since the default
setting is 100, according to google java style guide)

wiarlawd:
It looks like you're deleting trailing whitespace on unmodified lines.

Brett and/or PJ can finish this while I'm out. The library is slated to go out possibly by next Friday.

John L

test/com/google/enterprise/adaptor/HeartbeatHandlerTest.java
```diff
@@ -272,28 +270,32 @@ public void getDocContent(Request request, Response response)
...
-        .setAdaptor(mockAdaptor)
-        .setAuthzAuthority(mockAdaptor)
-        .setWatchdog(watchdog)
-        .setHeaderTimeoutMillis(100)
-        .setContentTimeoutMillis(100)
-        .build();
-    HeartbeatHandler handler = createHeartbeatHandlerBuilder()
-        .setWatchdog(watchdog2).setDocHandler(docHandler)
-        .build();
+    mockAdaptor =
```
BrettMichaelJohnson: This is why John is a stickler for arbitrary whitespace changes.  It took me far too long to detect that only the timeouts have changed.

BrettMichaelJohnson (dismissed):
We are planning to release the library this week, so we would like to wrap up this CL for inclusion.

I don't think John's suggestion re temp files applies here, since the test is not writing to the filesystem. I do agree that an explanation as to switching to locating resources is needed.  I suspect it is because Eclipse isn't setting the current directory when running JUnit.

PapaJoltOmega:
Spoke with szab100. Plans to update CL in next couple weeks. There is no need to have this CL in next cut.

PapaJoltOmega:
Hi Szabi, can you split of the heart-beat handler test piece.  It's flaky for me right now. Please send as seperate code review for more immediate submission. 

This PR has multiple orthogonal items in it.  Let's submit in peaces please.

szab100:
Yes, sure!

szab100:
@pjogooglecom @wiarlawd @BrettMichaelJohnson 
I just reformatted these changes (eclipse build xml + refactor test resource loading), pls review it again..

test/com/google/enterprise/adaptor/ApplicationTest.java
```diff
@@ -251,6 +252,15 @@ public void testBasicHttpsListen() throws Exception {
     thrown.expect(java.io.FileNotFoundException.class);
     conn.getContent();
   }
+  
+  private void initSSLKeystores() {
```
PapaJoltOmega: is this section simply redundant for non-eclipse users? why don't non-ecloipse users need it?  is there something that can be removed for non-eclipse users? some build.xml lines that are no longer necessary for unit testing?

PapaJoltOmega (commented):
(1) Flaky test fix was removed from this pull-request.  Can you updatge title of this commit to be only "Better Eclipse Support".  

(2) Can you run this setup without Eclipse to confirm that everythiong works for non-eclipse user and confirm here.

Thank you

szab100:
@pjogooglecom 

1) Done
2) Done - all tests pass with 'ant test' and from Eclipse

Re: initSSLKeyStore() - Yes, this was clearly redundant with the following:
 1) build.xml's "test" target, so I removed it from here (comments retained in TestHelper)
 2) GsaFeedFileSenderTest, so I moved the method to TestHelper and hooked it from both tests

wiarlawd:
LGTM

This conflicts with #28, so whoever goes second should pay attention.

Once PJ has approved this, please use "Squash and merge" rather than just clicking the "Merge pull request" button, which has a default of "Create a merge commit".

John L

PapaJoltOmega (approved):
Thank you

PR #19, opened 2017-03-29
Fix b/33924233 Use noIndex instead of notFound when skipping docs

BrettMichaelJohnson:
This change has a TransmissionDecision of do-not-index set
the "X-Robots-Tag=noindex" header, rather than returning a 404
response code.

### DIFF
--- a/src/com/google/enterprise/adaptor/DocumentHandler.java
+++ b/src/com/google/enterprise/adaptor/DocumentHandler.java
@@ -863,24 +863,22 @@ class DocumentHandler implements HttpHandler {
           log.log(Level.INFO, "changed SEND_BODY to DO_NOT_INDEX {0}",
               docId.getUniqueId());
           // Not using startSending. Instead we stop header timer, start
-          // content timer, send not-found page, and setup a sink for
+          // content timer, discard metadata and content, and setup a sink for
           // bytes provided by adaptor instance itself.
           watchdog.processingCompleted(workingThread);
           watchdog.processingStarting(workingThread, contentTimeoutMillis);
-          // Follow the precedent of HEAD doing a short-circuit out
-          // and potentially not reporting some later adaptor errors.
-          os = new CloseNotifyOutputStream(new SinkOutputStream());
-          // Send no metadata, other than Robots noindex.
+          // Send no body and no metadata, other than Robots noindex.
           ex.getResponseHeaders().add("X-Robots-Tag", "noindex");
-          // Do not wait until call to complete() in order to short-circuit.
-          // The following line is a 200 with an empty body.
-          HttpExchanges.respond(ex, HttpURLConnection.HTTP_OK,
-              "text/plain", new byte[0]);
+          int rc = HttpURLConnection.HTTP_OK;
+          HttpExchanges.startResponse(ex, rc, "text/plain", /*hasBody=*/ false);
+          os = new SinkOutputStream(new CloseNotifyOutputStream(
+              ex.getResponseBody()));
         } else if (state == State.SEND_BODY_TRANSFORMED_TO_HEAD) {
           log.log(Level.INFO, "changed SEND_BODY to HEAD {0}",
               docId.getUniqueId());
           startSending(true);  // Lie about content. Will be 0bytes chunked.
-          os = new CloseNotifyOutputStream(new SinkOutputStream());
+          os = new SinkOutputStream(new CloseNotifyOutputStream(
+              ex.getResponseBody()));
         } else {
           throw new IllegalStateException("unexpected state: " + state);
         }
--- a/src/com/google/enterprise/adaptor/DocumentHandler.java
+++ b/src/com/google/enterprise/adaptor/DocumentHandler.java
@@ -679,13 +679,22 @@ class DocumentHandler implements HttpHandler {
         Note: short-circuits closing connection and so does not propagate
         errors that may happen related to document contents. */
     HEAD,
+    /** 200 response code; do NOT send headers (other than Robots noindex).
+        Note: short-circuits closing connection and so does not propagate
+        errors that may happen related to document contents. */
+    HEAD_TRANSFORMED_TO_NO_INDEX,
     /** Stream document bytes as adaptor gives them. */
     SEND_BODY,
-    /** 200 response code; do send headers; drop document body; use same 
-        timeouts and error handling as SEND_BODY. */
+    /** 404 response code with same timeouts and error handling as SEND_BODY. */
+    SEND_BODY_TRANSFORMED_TO_NO_INDEX,
+    /** 200 response code; do send headers (other than Robots noindex);
+        drop document body; use same timeouts and error handling as SEND_BODY.*/
     SEND_BODY_TRANSFORMED_TO_HEAD,
     /** No file content to send but we send metadata and ACLs. */
     NO_CONTENT,
+    /** 204 or 304 response code; no metadata (other than Robots noindex);
+        no ACLs. */
+    NO_CONTENT_TRANSFORMED_TO_NO_INDEX
   }
 
   /**
@@ -766,6 +775,18 @@ class DocumentHandler implements HttpHandler {
       }
       if (state == State.NO_CONTENT) {
         startSending(false);
+      } else if (state == State.NO_CONTENT_TRANSFORMED_TO_NO_INDEX) {
+        log.log(Level.INFO, "changed NO_CONTENT to DO_NOT_INDEX {0}",
+            docId.getUniqueId());
+        // Send no metadata, other than Robots noindex.
+        ex.getResponseHeaders().add("X-Robots-Tag", "noindex");
+        // Respond with 304 instead of 204 when talking with non GSA requests
+        // such as browsers.
+        HttpExchanges.respond(ex,
+            (requestIsFromFullyTrustedClient(ex)
+                ? HttpURLConnection.HTTP_NO_CONTENT
+                : HttpURLConnection.HTTP_NOT_MODIFIED),
+            "text/plain", (byte[]) null);
       } else {
         throw new IllegalStateException("unexpected state: " + state);
       }
@@ -778,7 +799,9 @@ class DocumentHandler implements HttpHandler {
           // We will need to make an OutputStream.
           break;
         case HEAD:
+        case HEAD_TRANSFORMED_TO_NO_INDEX:
         case SEND_BODY:
+        case SEND_BODY_TRANSFORMED_TO_NO_INDEX:
         case SEND_BODY_TRANSFORMED_TO_HEAD:
           // Already called before. Provide saved OutputStream.
           return os;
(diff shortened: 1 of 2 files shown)


wiarlawd:
The changes to the state machine make me nervous, for two reasons. First, I don't really understand it. Second, we decided to not use setNoFollow, on a nudge, mainly because it meant doing nothing with some slight benefit. However, seeing this change, and the comments in the state engine, makes me think that when we use setNoIndex like this, we should also avoid sending metadata, ACLs, and content. That is more consistent with one of the main benefits of SkipDocumentFilter: do no bother sending things to the GSA that you don't want to index, and retains the main benefits of this change, which is simply to change the message in Index Diagnostics. I don't mind losing followed links as a result, which was a very minor point.

John L

src/com/google/enterprise/adaptor/DocumentHandler.java
```diff
@@ -847,30 +859,28 @@ public OutputStream getOutputStream() throws IOException {
                   .createPipeline(os, originalContentType, metadata);
             }
           }
-        } else if (state == State.SEND_BODY_TRANSFORMED_TO_NOT_FOUND) {
-          log.log(Level.INFO, "changed SEND_BODY to NOT_FOUND {0}",
+        } else if (state == State.SEND_BODY_TRANSFORMED_TO_NO_INDEX) {
+          log.log(Level.INFO, "changed SEND_BODY to DO_NOT_INDEX {0}",
               docId.getUniqueId());
           // Not using startSending. Instead we stop header timer, start
           // content timer, send not-found page, and setup a sink for
```
wiarlawd: Update "not-found page".

test/com/google/enterprise/adaptor/DocumentHandlerTest.java
```diff
@@ -516,14 +516,14 @@ public void getDocContent(Request request, Response response)
         .build();
     mockAdaptor.documentBytes = new byte[] {1, 2, 3};
     handler.handle(ex);
-    assertEquals(404, ex.getResponseCode());
-    assertEquals(null,
-         ex.getResponseHeaders().getFirst("X-Gsa-External-Metadata"));
-    assertArrayEquals("Error 404: Not Found".getBytes(), ex.getResponseBytes());
+    assertEquals(200, ex.getResponseCode());
+    assertEquals("noindex", ex.getResponseHeaders().getFirst("X-Robots-Tag"));
+    assertNull(ex.getResponseHeaders().getFirst("X-Gsa-External-Metadata"));
```
wiarlawd: Revert to assertEquals as discussed.

test/com/google/enterprise/adaptor/DocumentHandlerTest.java
```diff
@@ -550,11 +550,12 @@ public void getDocContent(Request request, Response response)
...
     handler.handle(headEx);
-    assertEquals(404, headEx.getResponseCode());
-    assertEquals(null,
-         headEx.getResponseHeaders().getFirst("X-Gsa-External-Metadata"));
-    // 404 response to HEAD request does not have content (it is HEAD)
-    assertArrayEquals("".getBytes(), headEx.getResponseBytes());
+    assertEquals(200, headEx.getResponseCode());
+    assertEquals("noindex",
+        headEx.getResponseHeaders().getFirst("X-Robots-Tag"));
+    assertNull(headEx.getResponseHeaders().getFirst("X-Gsa-External-Metadata"));
```
wiarlawd: ditto

src/com/google/enterprise/adaptor/DocumentHandler.java
```diff
@@ -847,30 +859,28 @@ public OutputStream getOutputStream() throws IOException {
...
-          countingOs = new CountingOutputStream(new CloseNotifyOutputStream(
-              ex.getResponseBody()));
-          log.finest("before writing chunked not-found response");
-          countingOs.write(
-              Translation.HTTP_NOT_FOUND.toString().getBytes(ENCODING));
-          log.finest("after writing chunked not-found response");
-          os = new SinkOutputStream(countingOs);
+          // Follow the precedent of HEAD doing a short-circuit out
+          // and potentially not reporting some later adaptor errors.
+          os = new CloseNotifyOutputStream(new SinkOutputStream());
```
wiarlawd: As discussed, preserve the call to ex.getResponseBody() as new SinkOS(new CloseNotifyOS(ex.getResponseBody())), both here and below.

wiarlawd (changes requested):
As discussed.

src/com/google/enterprise/adaptor/DocumentHandler.java
```diff
@@ -847,30 +859,26 @@ public OutputStream getOutputStream() throws IOException {
...
           // bytes provided by adaptor instance itself.
           watchdog.processingCompleted(workingThread);
           watchdog.processingStarting(workingThread, contentTimeoutMillis);
-          int rc = HttpURLConnection.HTTP_NOT_FOUND;
-          HttpExchanges.startResponse(ex, rc, "text/plain", /*hasBody=*/ true);
-          countingOs = new CountingOutputStream(new CloseNotifyOutputStream(
+          // Send no body and no metadata, other than Robots noindex.
+          ex.getResponseHeaders().add("X-Robots-Tag", "noindex");
+          int rc = HttpURLConnection.HTTP_OK;
+          HttpExchanges.startResponse(ex, rc, "text/plain", /*hasBody=*/ false);
```
wiarlawd, line 873: I don't really understand the point of lying and sending 0 bytes of chunked content below, but why not preserve that behavior here?

test/com/google/enterprise/adaptor/DocumentHandlerTest.java
```diff
@@ -725,11 +728,11 @@ public void getDocContent(Request request, Response response)
...
     mockAdaptor.documentBytes = new byte[] {1, 2, 3};
     handler.handle(ex);
-    assertEquals(404, ex.getResponseCode());
+    assertEquals(204, ex.getResponseCode());
+    assertEquals("noindex", ex.getResponseHeaders().getFirst("X-Robots-Tag"));
     assertEquals(null,
-         ex.getResponseHeaders().getFirst("X-Gsa-External-Metadata"));
-    // The 204 that would not have content becomes a 404 with content
-    assertArrayEquals("Error 404: Not Found".getBytes(), ex.getResponseBytes());
+        ex.getResponseHeaders().getFirst("X-Gsa-External-Metadata"));
```
wiarlawd: Maybe unfix broken indentation for git history?

wiarlawd (approved):
LGTM, with minor/trivial suggestions.

Please wait for PJ to review as well.

John L

src/com/google/enterprise/adaptor/DocumentHandler.java
```diff
@@ -847,30 +859,26 @@ public OutputStream getOutputStream() throws IOException {
...
-          countingOs.write(
-              Translation.HTTP_NOT_FOUND.toString().getBytes(ENCODING));
-          log.finest("after writing chunked not-found response");
-          os = new SinkOutputStream(countingOs);
         } else if (state == State.SEND_BODY_TRANSFORMED_TO_HEAD) {
           log.log(Level.INFO, "changed SEND_BODY to HEAD {0}",
               docId.getUniqueId());
           startSending(true);  // Lie about content. Will be 0bytes chunked.
-          countingOs = new CountingOutputStream(new CloseNotifyOutputStream(
+          os = new SinkOutputStream(new CloseNotifyOutputStream(
```
PapaJoltOmega, line 880: Why this change?

src/com/google/enterprise/adaptor/DocumentHandler.java
PapaJoltOmega: Chunked was intentional. I believe it was to preserve error behaviour because completed and successfully closed connection is different from one that results in 500 half way through document processing.

src/com/google/enterprise/adaptor/DocumentHandler.java
```diff
@@ -847,30 +859,26 @@ public OutputStream getOutputStream() throws IOException {
...
               docId.getUniqueId());
           // Not using startSending. Instead we stop header timer, start
-          // content timer, send not-found page, and setup a sink for
+          // content timer, discard metadata and content, and setup a sink for
           // bytes provided by adaptor instance itself.
           watchdog.processingCompleted(workingThread);
           watchdog.processingStarting(workingThread, contentTimeoutMillis);
-          int rc = HttpURLConnection.HTTP_NOT_FOUND;
-          HttpExchanges.startResponse(ex, rc, "text/plain", /*hasBody=*/ true);
-          countingOs = new CountingOutputStream(new CloseNotifyOutputStream(
```
PapaJoltOmega, line 860: Why this change?

PapaJoltOmega (commented):
Thank you for this CL. I was away last week and didn't get to review.  

Looks like this CL was submitted to master HEAD already and I'm not yet sure it is right. 

Thank you

src/com/google/enterprise/adaptor/DocumentHandler.java
BrettMichaelJohnson: The CountingOuputStream provided no benefit here (and in SEND_BODY_TRANSFORMED_TO_NO_INDEX above). It is used only to journal the number of bytes of content written.  In these two cases, that number is always 0 (since we are no longer forging a 404 response page).  If countingOs is null, then the journal records 0 bytes written.

src/com/google/enterprise/adaptor/DocumentHandler.java
BrettMichaelJohnson: So you concur with John, in that I should lie here with hasBody = true, even though no bytes get written?  I can do that.

PapaJoltOmega: Yes, because at this moment I believe the behaviour is different.

I think the commented behaviour on line 691 says that "use same timeouts and error handling as SEND_BODY.".  

On one hand we have: 0 sized chunked that reads entire stream before closing connection 

On other we have: 0 sized that is immediately closed

There are timing differences and there are differences when there is error in reading body.

src/com/google/enterprise/adaptor/DocumentHandler.java
PapaJoltOmega: I would appreciate anything that can be done to limit this CL's modifications because changes to this file stress my mental capabilities.  Does your comment mean that removing the COuntingOutputStream wrapper is not essential to this CL and that we can consider such removal orthogonally to this CL?  I would greatly appreciate any fragment that can be extracted and seperated be done in seperate CL. Thank you

PapaJoltOmega (commented):
Thank you

src/com/google/enterprise/adaptor/DocumentHandler.java
wiarlawd: Fwiw, Brett removed CountingOutputStream for basically the same reason: this code is hard to reason about, with many subtle differences, and having no-op code wasn't helping us understand the subtle differences. He's going to put up a new commit here to minimize diffs, but it will look a little strange on master (assuming there are other subsequent changes), as the countingOs will be removed, added back, and then possibly removed again.

PapaJoltOmega (commented):
LGTM.  Thank you.

BrettMichaelJohnson:
This pull request was reverted, and a new one with additional changes was created [here](https://github.com/googlegsa/library/pull/22)

PR #20, opened 2017-04-03
Fix b/21168971 Use a separate formatter without color for FileHandler

BrettMichaelJohnson:


logging.properties
```diff
@@ -9,7 +9,7 @@ java.util.logging.ConsoleHandler.formatter = com.google.enterprise.adaptor.Custo
 # then uncomment the following line to force colors off.
 # com.google.enterprise.adaptor.CustomFormatter.useColor = false
 
-java.util.logging.FileHandler.formatter=com.google.enterprise.adaptor.CustomFormatter
+java.util.logging.FileHandler.formatter=com.google.enterprise.adaptor.CustomFormatter.NoColor
```
wiarlawd: The choice between ".useColor = false" and ".NoColor" is confusing, since they both accomplish the same thing. Maybe change the comments to suggest using the NoColor formatter instead of useColor = false? We could even add a parallel Color subclass with setColor(true), and use the two new classes in preference to the static property, while preserving the original class and property for backward compatibility.

src/com/google/enterprise/adaptor/CustomFormatter.java
```diff
@@ -65,6 +65,16 @@ public CustomFormatter() {
     }
   }
 
+  public CustomFormatter(boolean useColor) {
+    setUseColor(useColor);
+  }
+
+  public class NoColor extends CustomFormatter {
```
wiarlawd: static?

wiarlawd: Add javadoc.

src/com/google/enterprise/adaptor/CustomFormatter.java
```diff
@@ -65,6 +65,16 @@ public CustomFormatter() {
     }
   }
 
+  public CustomFormatter(boolean useColor) {
```
wiarlawd: private?

wiarlawd (commented):
More comments than you might have thought possible. ;-)

John L

logging.properties
BrettMichaelJohnson: Done.

src/com/google/enterprise/adaptor/CustomFormatter.java
BrettMichaelJohnson: Done.

src/com/google/enterprise/adaptor/CustomFormatter.java
BrettMichaelJohnson: Done.

BrettMichaelJohnson: Done.

src/com/google/enterprise/adaptor/CustomFormatter.java
```diff
@@ -65,6 +65,24 @@ public CustomFormatter() {
...
   }
 
+  private CustomFormatter(boolean useColor) {
+    setUseColor(useColor);
+  }
+
+  /** A CustomFormatter that forces the use of colors. */
+  public static class Color extends CustomFormatter {
+    public Color() {
+      super(true);
```
wiarlawd, line 75: This isn't what I meant, though it's what I wrote, and I like this better. I just meant ignore the useColor property, rather than bypassing the auto-detect as well. But this is basically Color/NoColor is a full replacement for useColor, while the base CustomFormatter implements the auto-detect, and is not deprecated. SGTM.

src/com/google/enterprise/adaptor/CustomFormatter.java
```diff
@@ -65,6 +65,24 @@ public CustomFormatter() {
     }
```
wiarlawd, line 65: If we're deprecating useColor, then we should log a message about it here, and update the class javadoc to mention Color and NoColor instead of useColor, or at least in preference to it.

PapaJoltOmega (approved):
LGTm. Thank you

PapaJoltOmega (changes requested):
What is the behaviour when someone upgrades their connector?  Does their connector continue to run?  Did we make some config files invalid and cause connector to fail startup?

BrettMichaelJohnson:
The CustomFormatter class remains public and largely unchanged (including looking for useColor config), so existing logging.properties files will still work.  The installers don't really "upgrade".  If you point it at an existing installation, Windows prompts you when overwriting existing files, so the user can choose to use their existing logging.properties or overwrite it with the new one.  The logging.properties files used by each installer are not not the ones in GitHub for each connector (or the library's).  They are currently in a set of input files used for each connector's installer.  They are all nearly identical, except for setting the logging level for each adaptor's package, and the name of log files.

PapaJoltOmega (approved):
Thank you.  Happy to hear confirmation that  admin's existing logging.properties continues to be valid.

PR #21, opened 2017-04-05
Revert "Fix b/33924233 Use noIndex instead of notFound when skipping docs"

BrettMichaelJohnson:
Reverts googlegsa/library#19


PapaJoltOmega (approved):
LGTM. Thank you

PR #22, opened 2017-04-05
Fix b/33924233 Use noIndex instead of notFound when skipping docs

BrettMichaelJohnson:
This change has a TransmissionDecision of do-not-index set
the "X-Robots-Tag=noindex" header, then discard metadata,
ACLs, and content; rather than returning a 404 response code.

### DIFF
--- a/src/com/google/enterprise/adaptor/DocumentHandler.java
+++ b/src/com/google/enterprise/adaptor/DocumentHandler.java
@@ -863,24 +863,22 @@ class DocumentHandler implements HttpHandler {
           log.log(Level.INFO, "changed SEND_BODY to DO_NOT_INDEX {0}",
               docId.getUniqueId());
           // Not using startSending. Instead we stop header timer, start
-          // content timer, send not-found page, and setup a sink for
+          // content timer, discard metadata and content, and setup a sink for
           // bytes provided by adaptor instance itself.
           watchdog.processingCompleted(workingThread);
           watchdog.processingStarting(workingThread, contentTimeoutMillis);
-          // Follow the precedent of HEAD doing a short-circuit out
-          // and potentially not reporting some later adaptor errors.
-          os = new CloseNotifyOutputStream(new SinkOutputStream());
-          // Send no metadata, other than Robots noindex.
+          // Send no body and no metadata, other than Robots noindex.
           ex.getResponseHeaders().add("X-Robots-Tag", "noindex");
-          // Do not wait until call to complete() in order to short-circuit.
-          // The following line is a 200 with an empty body.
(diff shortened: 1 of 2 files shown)


src/com/google/enterprise/adaptor/DocumentHandler.java
```diff
@@ -774,9 +775,18 @@ public void respondNoContent() throws IOException{
...
             docId.getUniqueId());
+        // Send no metadata, other than Robots noindex.
+        ex.getResponseHeaders().add("X-Robots-Tag", "noindex");
+        // Respond with 304 instead of 204 when talking with non GSA requests
+        // such as browsers.
+        HttpExchanges.respond(ex,
+            (requestIsFromFullyTrustedClient(ex)
+                ? HttpURLConnection.HTTP_NO_CONTENT
+                : HttpURLConnection.HTTP_NOT_MODIFIED),
+            "text/plain", (byte[]) null);
```
wiarlawd: This isn't a HEAD request, so perhaps we should follow the chunked examples elsewhere and use new byte[0] here instead of null. Since getOutputStream won't be called here, or it is an error if it is, I'm not sure whether we will get bitten by the double-close exception in HttpsExchanges.respond when the response is non-null, or not.

wiarlawd (approved):
LGTM

Please wait for PJ.

John L

PapaJoltOmega (commented):
I've spent lots of time thinking about this CL.

I hvae two concerns:
(A) I don't have confidence in GSA's robots behaviour like I do in GSA's 404 behaviour.
(B) There is a security difference as jlacey@ points out to someone that used 404 intentionally and now depends on it.

The only benefit that I know of that we get is a cleaner message in administration console.  It seems to me like the benefit does not outweigh the risk for the GSA, especially during product shutdown.

BrettMichaelJohnson:
Why are you not confident in the GSA handling of noIndex? I tested it with both unindexed and previously indexed content and it behaves as desired for this feature.  If you were unsure of its use, why did you approve this change in the FS adaptor.  We are planning the same change in most of the other adaptors, as well.

BrettMichaelJohnson:
As far as  the intentional use of 404, would you like me to add another transform decision of "not found"?

PapaJoltOmega:
We're in a situation where the signal to not index has been implemented as 404 and that's working out well.  


I';m not sure if the suggestion is to have two decisions NOT_FOUND and NO_INDEX that will both produce 404, or to have those two decisions have different implementations where folks that are cuirrently using NO_INDEX would have to switch to NOT_FOUND in order to retain current behavior.

I think in both cases, not necessary.  

Thank you

BrettMichaelJohnson:
Abandoning change.

PR #23, opened 2017-04-07
Fix b/32321298 Part1: SkipDocumentFilter conditionally sets TransmissionDecision

BrettMichaelJohnson:
This change changes SkipDocumentFilter to only set the TransmissionDecision
to do-not-index when its conditions are met. Otherwise, the TransmissionDecision
remains unchanged. This allows several SkipDocumentFilters to be deployed in
the metadata transform pipeline.

### DIFF
--- a/src/com/google/enterprise/adaptor/prebuilt/SkipDocumentFilter.java
+++ b/src/com/google/enterprise/adaptor/prebuilt/SkipDocumentFilter.java
@@ -169,9 +169,8 @@ public class SkipDocumentFilter implements MetadataTransform {
   }
 
   /**
-   * Adds a single {@code Map.Entry} to the {@code params Map}: key
-   * {@code Transmission-Decision}, value {@code TransmissionDecision.AS_IS} to
-   * indicate that the document is to be retained, or value
+   * Conditionally adds a single {@code Map.Entry} to the {@code params Map}:
+   * key {@code Transmission-Decision}, value
    * {@code TransmissionDecision.DO_NOT_INDEX} to indicate that the document is
    * to be skipped.  The decision is based on settings of the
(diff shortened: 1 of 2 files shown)


src/com/google/enterprise/adaptor/prebuilt/SkipDocumentFilter.java
```diff
@@ -203,19 +202,9 @@ public void transform(Metadata metadata, Map<String, String> params) {
             + corpora);
         params.put(MetadataTransform.KEY_TRANSMISSION_DECISION,
             TransmissionDecision.DO_NOT_INDEX.toString());
-      } else {
-        log.info("Not skipping document " + docId + ", because we did not find "
```
wiarlawd: Why did you choose not to preserve the log messages? What about keeping them at FINE or lower?

BrettMichaelJohnson: Done.

wiarlawd (approved):
LGTM

Please wait for PJ to review.

John L

PapaJoltOmega (commented):
Do we have plans to change the AS_IS behaviour of FilterMimetypes?

Is AS_IS useful in any circumstance?

BrettMichaelJohnson:
The change to FilterMimetypes is a separate CL. Since SkipDocumentFilter is limited to using RegEx to make a decision, AS_IS could be used by a subsequent filter to restore inadvertently skipped content (John used forcing folders to the be indexed for graph traversers as an example.)  The current plan is to adapt SkipDocumentFilter into a RegExDecisionFilter with a configurable decision property. Then SkipDocumentFilter would become a trivial subclass with a decision of "do-not-index".

PapaJoltOmega (approved):

PR #24, opened 2017-04-07
Fix b/32321298 Part2: FilterMimetypes conditionally sets TransmissionDecision

BrettMichaelJohnson:
This change changes FilterMimetypes to only set the TransmissionDecision
to do-not-index or do-not-index-content when its conditions are met.
Otherwise, the TransmissionDecision remains unchanged.
This allows FilterMimetyps to be used with other TransmissionDecision
filters in the metadata transform pipeline.

### DIFF
--- a/src/com/google/enterprise/adaptor/prebuilt/FilterMimetypes.java
+++ b/src/com/google/enterprise/adaptor/prebuilt/FilterMimetypes.java
@@ -129,6 +129,7 @@ public class FilterMimetypes implements MetadataTransform {
       ct = ct.substring(0, semicolonIndex);
     }
     ct = ct.trim().toLowerCase();
+    // First look in the cache to see if we have encounted this mimetype before.
     String decision = lookupDecision(ct);
     if (null != decision) {
       if (!decision.equals(TransmissionDecision.AS_IS.toString())) {
@@ -138,6 +139,8 @@ public class FilterMimetypes implements MetadataTransform {
     }
     if (supportedExplicit.contains(ct)) {
       log.log(Level.FINE, ct + "is explicitly supported");
+      // Even though we do not explicitly set the decision, still cache
+      // the result for future encounters of this same mime-type.
       insertDecision(ct, TransmissionDecision.AS_IS.toString());
     } else if (unsupportedExplicit.contains(ct)) {
       log.log(Level.FINE, ct + "is explicitly unsupported");
--- a/src/com/google/enterprise/adaptor/prebuilt/FilterMimetypes.java
+++ b/src/com/google/enterprise/adaptor/prebuilt/FilterMimetypes.java
@@ -132,7 +132,7 @@ public class FilterMimetypes implements MetadataTransform {
     String decision = lookupDecision(ct);
     if (null != decision) {
       if (!decision.equals(TransmissionDecision.AS_IS.toString())) {
-        params.put("Transmission-Decision", decision);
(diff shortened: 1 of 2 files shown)


wiarlawd (approved):
LGTM

The changes here seem fine, but the logging was already inconsistent internally (we don't even always log skipped documents), and inconsistent with SkipDocumentFilter. Maybe pick some approach to use consistently? Could be a separate CL.

John L

src/com/google/enterprise/adaptor/prebuilt/FilterMimetypes.java
```diff
@@ -137,8 +137,6 @@ public void transform(Metadata metadata, Map<String, String> params) {
     if (supportedExplicit.contains(ct)) {
       log.log(Level.FINE, ct + "is explicitly supported");
       insertDecision(ct, TransmissionDecision.AS_IS.toString());
```
PapaJoltOmega: I think we will be responding with explicit AS_IS for everything but first instance of a type.

src/com/google/enterprise/adaptor/prebuilt/FilterMimetypes.java
```diff
@@ -151,8 +149,6 @@ public void transform(Metadata metadata, Map<String, String> params) {
           TransmissionDecision.DO_NOT_INDEX.toString());
     } else if (matches(supportedGlobs, ct, "supported by glob")) {
       insertDecision(ct, TransmissionDecision.AS_IS.toString());
```
PapaJoltOmega, line 156: ditto

PapaJoltOmega (commented):
Ignore question regarding FilterMimetypes from last CL :)

src/com/google/enterprise/adaptor/prebuilt/FilterMimetypes.java
```diff
@@ -131,14 +131,14 @@ public void transform(Metadata metadata, Map<String, String> params) {
     ct = ct.trim().toLowerCase();
     String decision = lookupDecision(ct);
     if (null != decision) {
-      params.put("Transmission-Decision", decision);
+      if (!decision.equals(TransmissionDecision.AS_IS.toString())) {
+        params.put("Transmission-Decision", decision);
```
wiarlawd: While we're reindenting this line, maybe use MetadataTransform.KEY_TRANSMISSION_DECISION as the deleted lines below did?

src/com/google/enterprise/adaptor/prebuilt/FilterMimetypes.java
```diff
@@ -131,14 +131,14 @@ public void transform(Metadata metadata, Map<String, String> params) {
     ct = ct.trim().toLowerCase();
     String decision = lookupDecision(ct);
     if (null != decision) {
-      params.put("Transmission-Decision", decision);
+      if (!decision.equals(TransmissionDecision.AS_IS.toString())) {
```
PapaJoltOmega: Please add note that we store AS_IS decisions for short circuiting search, however we do provide the decision back because we don't want to override another filters exclusion (that's the reason right?).

PapaJoltOmega (approved):

src/com/google/enterprise/adaptor/prebuilt/FilterMimetypes.java
BrettMichaelJohnson: Done.

PR #25, opened 2017-04-07
Fix b/29451090 adding validation and tests for timeout parameters

mpaggi:
Adding code and tests to check the values of the following properties:
adaptor.docContentTimeoutSecs
adaptor.docHeaderTimeoutSecs
adaptor.heartbeatTimeoutSecs
In case the value set in config file is invalid, default value is returned and a warning is logged.

### DIFF
--- a/src/com/google/enterprise/adaptor/Config.java
+++ b/src/com/google/enterprise/adaptor/Config.java
@@ -650,7 +650,7 @@ public class Config {
         return Long.parseLong(secondsAsString) * 1000;
       } catch (NumberFormatException nfe) {
         throw new InvalidConfigurationException("Invalid value for "
-            + property + ". Only a numeric value is accepted.");
+            + property + ". Only a positive integer value is accepted.");
       }
     }
   }
--- a/src/com/google/enterprise/adaptor/Config.java
+++ b/src/com/google/enterprise/adaptor/Config.java
@@ -643,19 +643,19 @@ public class Config {
     String secondsAsString = getValue(property).trim();
     if ("adaptor.heartbeatTimeoutSecs".equalsIgnoreCase(property) &&
         "".equals(secondsAsString)) {
-      // if heartbeatTimeoutSecs is empty, default value is docHeaderTimeoutSecs value
+      // if heartbeatTimeoutSecs is empty, default value is docHeaderTimeoutSecs
       return getAdaptorDocHeaderTimeoutMillis();
     } else {
       if ("0".equals(secondsAsString) || "".equals(secondsAsString) ||
           secondsAsString.startsWith("-") ) {
-        throw new InvalidConfigurationException("Invalid value for " + property +
-            ". Zero, empty and negative values are not accepted.");
+        throw new InvalidConfigurationException("Invalid value for " + property
+            + ". Zero, empty and negative values are not accepted.");
       } else {
         try {
           return Long.parseLong(secondsAsString) * 1000;
         } catch (NumberFormatException nfe) {
-          throw new InvalidConfigurationException("Invalid value for " + property +
-              ". Only a numeric value is accepted.");
+          throw new InvalidConfigurationException("Invalid value for "
+              + property + ". Only a numeric value is accepted.");
         }
       }
     }
--- a/src/com/google/enterprise/adaptor/Config.java
+++ b/src/com/google/enterprise/adaptor/Config.java
@@ -639,44 +639,42 @@ public class Config {
     return getValue("adaptor.fullListingSchedule");
   }
 
-  long getAdaptorIncrementalPollPeriodMillis() {
-    return Long.parseLong(getValue("adaptor.incrementalPollPeriodSecs")) * 1000;
+  public long validateTimeout(String property) {
+    String secondsAsString = getValue(property).trim();
+    if ("adaptor.heartbeatTimeoutSecs".equalsIgnoreCase(property) &&
+        "".equals(secondsAsString)) {
+      // if heartbeatTimeoutSecs is empty, default value is docHeaderTimeoutSecs value
+      return getAdaptorDocHeaderTimeoutMillis();
+    } else {
+      if ("0".equals(secondsAsString) || "".equals(secondsAsString) ||
+          secondsAsString.startsWith("-") ) {
+        throw new InvalidConfigurationException("Invalid value for " + property +
+            ". Zero, empty and negative values are not accepted.");
+      } else {
+        try {
+          return Long.parseLong(secondsAsString) * 1000;
+        } catch (NumberFormatException nfe) {
+          throw new InvalidConfigurationException("Invalid value for " + property +
+              ". Only a numeric value is accepted.");
+        }
+      }
+    }
   }
 
-  public long validateTimeout(String property, String defaultValue,
-      boolean allowNeg, boolean allowEmpty) {
-    String secondsAsString = getValue(property).trim();
-    try {
-      if ((null == secondsAsString) || "0".equals(secondsAsString) ||
-          (("".equals(secondsAsString) && !(allowEmpty))) ||
-          (secondsAsString.startsWith("-") && !(allowNeg))) {
-        log.log(Level.WARNING, "Invalid value for " + property
-                + ", it has been set to default " + defaultValue + " seconds.");
-      } else {
-          return Long.parseLong(secondsAsString) * 1000;
-      }
-    } catch (NumberFormatException e) {
-      log.log(Level.WARNING, "Invalid value for " + property
-              + ", it has been set to default " + defaultValue + " seconds.");
-    }
-    // For invalid value, returning default
-    // We do not reach this line if value is valid
-    return Long.parseLong(defaultValue) * 1000;
+  long getAdaptorIncrementalPollPeriodMillis() {
+    return validateTimeout("adaptor.incrementalPollPeriodSecs");
   }
 
   long getAdaptorDocHeaderTimeoutMillis() {
-    return validateTimeout("adaptor.docHeaderTimeoutSecs", "30", false, false);
+    return validateTimeout("adaptor.docHeaderTimeoutSecs");
   }
 
   long getAdaptorDocContentTimeoutMillis() {
-    return validateTimeout("adaptor.docContentTimeoutSecs", "180",
-                           false, false);
+    return validateTimeout("adaptor.docContentTimeoutSecs");
   }
 
   long getAdaptorHeartbeatTimeoutMillis() {
-    return validateTimeout("adaptor.heartbeatTimeoutSecs", Long.toString(
-                           getAdaptorDocHeaderTimeoutMillis() / 1000),
-                           false, false);
+    return validateTimeout("adaptor.heartbeatTimeoutSecs");
   }
 
   /**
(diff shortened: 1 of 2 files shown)


src/com/google/enterprise/adaptor/Config.java
```diff
@@ -643,20 +643,40 @@ long getAdaptorIncrementalPollPeriodMillis() {
     return Long.parseLong(getValue("adaptor.incrementalPollPeriodSecs")) * 1000;
   }
 
+  public long validateTimeout(String property, String defaultValue,
+      boolean allowNeg, boolean allowEmpty) {
+    String secondsAsString = getValue(property).trim();
+    try {
+      if ((null == secondsAsString) || "0".equals(secondsAsString) ||
```
BrettMichaelJohnson: Config.getValue() does not return null.

src/com/google/enterprise/adaptor/Config.java
```diff
@@ -643,20 +643,40 @@ long getAdaptorIncrementalPollPeriodMillis() {
     return Long.parseLong(getValue("adaptor.incrementalPollPeriodSecs")) * 1000;
   }
 
+  public long validateTimeout(String property, String defaultValue,
+      boolean allowNeg, boolean allowEmpty) {
```
PapaJoltOmega: Why do we have these last two booleans if they are always passed in false?

src/com/google/enterprise/adaptor/Config.java
```diff
@@ -643,20 +643,40 @@ long getAdaptorIncrementalPollPeriodMillis() {
...
+        log.log(Level.WARNING, "Invalid value for " + property
+                + ", it has been set to default " + defaultValue + " seconds.");
+      } else {
+          return Long.parseLong(secondsAsString) * 1000;
+      }
+    } catch (NumberFormatException e) {
+      log.log(Level.WARNING, "Invalid value for " + property
+              + ", it has been set to default " + defaultValue + " seconds.");
+    }
+    // For invalid value, returning default
```
PapaJoltOmega: Why do we decide to continue with default rather than die?

PapaJoltOmega (commented):
Thank you

src/com/google/enterprise/adaptor/Config.java
```diff
@@ -643,20 +643,40 @@ long getAdaptorIncrementalPollPeriodMillis() {
...
+      boolean allowNeg, boolean allowEmpty) {
+    String secondsAsString = getValue(property).trim();
+    try {
+      if ((null == secondsAsString) || "0".equals(secondsAsString) ||
+          (("".equals(secondsAsString) && !(allowEmpty))) ||
+          (secondsAsString.startsWith("-") && !(allowNeg))) {
+        log.log(Level.WARNING, "Invalid value for " + property
+                + ", it has been set to default " + defaultValue + " seconds.");
+      } else {
+          return Long.parseLong(secondsAsString) * 1000;
```
BrettMichaelJohnson: The pedantic mission-to-mars programmer would ensure that multiplying by 1000 will not overflow.

src/com/google/enterprise/adaptor/Config.java
```diff
@@ -643,20 +643,40 @@ long getAdaptorIncrementalPollPeriodMillis() {
...
+              + ", it has been set to default " + defaultValue + " seconds.");
+    }
+    // For invalid value, returning default
+    // We do not reach this line if value is valid
+    return Long.parseLong(defaultValue) * 1000;
+  }
+
   long getAdaptorDocHeaderTimeoutMillis() {
-    return Long.parseLong(getValue("adaptor.docHeaderTimeoutSecs")) * 1000;
+    return validateTimeout("adaptor.docHeaderTimeoutSecs", "30", false, false);
```
BrettMichaelJohnson: You are restating the default values supplied in the constructor here. I don't like that for 2 reasons:  
1) If the defaults are changed in the future, the change may be missed here.  
2) The default value may have been overridden by the connector using overrideKey(). (The FS adaptor overrides these values)

One solution might be to pull the default values from defaultConfig, but you need to be wary of nulls and computed values (not the case for these).

src/com/google/enterprise/adaptor/Config.java
```diff
@@ -643,20 +643,40 @@ long getAdaptorIncrementalPollPeriodMillis() {
     return Long.parseLong(getValue("adaptor.incrementalPollPeriodSecs")) * 1000;
```
BrettMichaelJohnson: you might as well validate this one too.

test/com/google/enterprise/adaptor/ConfigTest.java
```diff
@@ -330,4 +330,48 @@ public void testPropertiesParseEscapeASlash6() throws Exception {
...
+    // docheartbeatTimeoutSecs=0 and docheartbeatTimeoutSecs=-15 are invalid
+    // it should use the default value for adaptor.docHeaderTimeoutSecs
+    // that is 30 seconds.
+    String putInConfig = " gsa.hostname=not_used\n"
+                         + "adaptor.heartbeatTimeoutSecs=";
+    String invalidValuesToVerify[] = {"0", "-15", "", "NotValidValue"};
+    for (int  i = 0; i < invalidValuesToVerify.length; i++) {
+      configFile.setFileContents(putInConfig + invalidValuesToVerify[i]);
+      config.load(configFile);
+      assertEquals(Long.parseLong("30000"),
```
BrettMichaelJohnson: Consider using TimeUnit.SECONDS.toMillis(30).

src/com/google/enterprise/adaptor/Config.java
```diff
@@ -643,20 +643,40 @@ long getAdaptorIncrementalPollPeriodMillis() {
...
 
   long getAdaptorHeartbeatTimeoutMillis() {
-    String secondsAsString = getValue("adaptor.heartbeatTimeoutSecs");
-    if ((null == secondsAsString) || "".equals(secondsAsString.trim())) {
-      secondsAsString = getValue("adaptor.docHeaderTimeoutSecs");
-    }
-    return Long.parseLong(secondsAsString) * 1000;
+    return validateTimeout("adaptor.heartbeatTimeoutSecs", Long.toString(
+                           getAdaptorDocHeaderTimeoutMillis() / 1000),
+                           false, false);
```
BrettMichaelJohnson: The default value of heartbeatTimeoutSecs is "", so allowEmpty should probably be true. But that will still generate a WARNING in the logs.

test/com/google/enterprise/adaptor/ConfigTest.java
```diff
@@ -330,4 +330,48 @@ public void testPropertiesParseEscapeASlash6() throws Exception {
     config.load(configFile);
     assertEquals("\\\t\\\f", config.getValue("slash"));
   }
+
```
BrettMichaelJohnson, line 335: Add tests that verify that valid timeout config values are returned.

BrettMichaelJohnson (changes requested):

src/com/google/enterprise/adaptor/Config.java
BrettMichaelJohnson: In fact, I would probably have passed in min and max allowed values (which also does away with allowNeg, which is never true anyway).

src/com/google/enterprise/adaptor/Config.java
BrettMichaelJohnson: I agree with PJ.  For truly invalid values, throw InvalidConfigurationException.  This also avoids the problems with the defaults below. Although the heartbeatTimeout will still have to fall back to headerTimeout gracefully.

BrettMichaelJohnson:
Michele, we are planning a release of the adaptor library this week.  Please consider finishing this up so that is may be included.

PapaJoltOmega:
Michele, would you mind if Brett took this CL over? Thank you

mpaggi:
Sorry I did not have time to work on this before. Here the changes suggested.
I changed also the tests since 'thrown.expect' does not seems to works in a for cycle.
No problem for me if Brett wants to take over this CL.

src/com/google/enterprise/adaptor/Config.java
```diff
@@ -639,24 +639,42 @@ String getAdaptorFullListingSchedule() {
     return getValue("adaptor.fullListingSchedule");
   }
 
+  public long validateTimeout(String property) {
+    String secondsAsString = getValue(property).trim();
+    if ("adaptor.heartbeatTimeoutSecs".equalsIgnoreCase(property) &&
```
BrettMichaelJohnson: This belongs in getAdaptorHeartbeatTimeoutMillis, not here.

test/com/google/enterprise/adaptor/ConfigTest.java
```diff
@@ -330,4 +332,107 @@ public void testPropertiesParseEscapeASlash6() throws Exception {
...
+        config.getAdaptorHeartbeatTimeoutMillis());
+  }
+
+  @Test
+  public void testPropertiesIncrementalPollPeriodSecs() throws Exception {
+    // incrementalPollPeriodSecs=0 and incrementalPollPeriodSecs=-15 are invalid
+    String putInConfig = " gsa.hostname=not_used\n"
+                         + "adaptor.incrementalPollPeriodSecs=";
+    String invalidValuesToVerify[] = {"0", "-15", "NotValidValue", "",
+                                      Long.toString(Long.MAX_VALUE + 1)};
```
BrettMichaelJohnson: Long.MAX_VALUE + 1 = Long.MIN_VALUE
What you want is (Long.MAX_VALUE/1000)+1 , which will overflow when converted to milliseconds.

test/com/google/enterprise/adaptor/ConfigTest.java
```diff
@@ -330,4 +332,107 @@ public void testPropertiesParseEscapeASlash6() throws Exception {
...
     assertEquals("\\\t\\\f", config.getValue("slash"));
   }
+
+  @Test
+  public void testValidValuesForTimeouts() throws Exception {
+    configFile.setFileContents("gsa.hostname=not_used\n"
+                               + "adaptor.incrementalPollPeriodSecs=1700\n"
+                               + "adaptor.docHeaderTimeoutSecs=50\n"
+                               + "adaptor.docContentTimeoutSecs=190\n"
+                               + "adaptor.heartbeatTimeoutSecs=\n");
```
BrettMichaelJohnson: In this test, explicitly set heartbeat, then add another test that verifies the fallback value.

BrettMichaelJohnson (changes requested):

BrettMichaelJohnson:
PJ,
I took over this CL, and made (admittedly my own) suggested changes.
Please review.

src/com/google/enterprise/adaptor/Config.java
```diff
@@ -639,24 +639,41 @@ String getAdaptorFullListingSchedule() {
...
+    if ("0".equals(secondsAsString) || "".equals(secondsAsString) ||
+        secondsAsString.startsWith("-") ) {
+      throw new InvalidConfigurationException("Invalid value for " + property
+          + ". Zero, empty and negative values are not accepted.");
+    } else {
+      try {
+        return Long.parseLong(secondsAsString) * 1000;
+      } catch (NumberFormatException nfe) {
+        throw new InvalidConfigurationException("Invalid value for "
+            + property + ". Only a numeric value is accepted.");
```
PapaJoltOmega: Can we say numeric that also conveys integer?

PapaJoltOmega (approved):
LGTM.  Thank you

BrettMichaelJohnson (approved):

PR #27, opened 2017-04-11
Fix b/11781292 Part1: Add support for non-incremental groups push

BrettMichaelJohnson:
This change starts with the proposed changes in
    https://codereview.appspot.com/311210043/
    and alters it as follows:
    1) Add an overload of DocIdPusher.pushGroupDefinitions() that takes
       a boolean incremental flag.
    2) That same overloaded method also takes an optional group source name
       that may be used as the feed's groupsource, rather than the feed.name
       config property.
    3) Remove the implicit selection of incremental v. full group feeds based
       upon the state of the journal.
    4) Improve the tests to ensure that incremental flag and groupSource were
       propagated to the FeedFileSender correctly.


src/com/google/enterprise/adaptor/DocIdPusher.java
```diff
@@ -184,6 +184,43 @@ public GroupPrincipal pushGroupDefinitions(
...
+   * @param incremental if {@code true} incremental update is done, otherwise
+   *        full replacement is done.
+   * @param groupSource
+   * @param handler for dealing with errors pushing
+   * @return {@code null} on success, otherwise the first GroupPrincipal to fail
+   * @throws InterruptedException if interrupted and no definitions were sent
+   */
+  public GroupPrincipal pushGroupDefinitions(
+      Map<GroupPrincipal, ? extends Collection<Principal>> defs,
+      boolean caseSensitive, boolean incremental, String groupSource,
```
wiarlawd: I wonder about creating two enums for these boolean parameters.

src/com/google/enterprise/adaptor/DocIdSender.java
```diff
@@ -356,7 +368,7 @@ public GroupPrincipal pushGroupDefinitions(
       try {
         log.info("sending groups to GSA host name: " + config.getGsaHostname());
         fileSender.sendGroups(feedSourceName,
-            groupsDefXml, config.isServerToUseCompression());
+            groupsDefXml, config.isServerToUseCompression(), incremental);
```
wiarlawd: This parameter order feels a little random, but so do the other orders. Moving incremental after feedSourceName feels slightly better to me (sort of the reverse of pushGroupDefinitions and its friends).

src/com/google/enterprise/adaptor/GsaFeedFileSender.java
```diff
@@ -257,12 +259,12 @@ public void sendFull(String datasource, String xmlString,
    * Groupsource name is limited to [a-zA-Z_][a-zA-Z0-9_-]*.
    */
   void sendGroups(String groupsource, String xmlString,
-      boolean useCompression) throws IOException {
+      boolean useCompression, boolean incremental) throws IOException {
```
wiarlawd: From this context, I definitely don't like the order. incremental before useCompression or xmlString would feel better to me. This order pairs (A, B, D) with (A, B, C, D), rather than (A, B, D, C).

test/com/google/enterprise/adaptor/AccumulatingDocIdPusher.java
```diff
@@ -60,8 +60,8 @@ public DocId pushNamedResources(Map<DocId, Acl> resources,
   @Override
   public GroupPrincipal pushGroupDefinitions(
       Map<GroupPrincipal, ? extends Collection<Principal>> defs,
-      boolean caseSensitive, ExceptionHandler handler)
-      throws InterruptedException {
+      boolean caseSensitive, boolean incremental, String sourceName,
+      ExceptionHandler exceptionHandler) throws InterruptedException {
```
wiarlawd: This could be a good time (in a separate CL) to look at sharing this class with the connectors, and possibly including a shared UnsupportedDocIdPusher (there's also DelegatingDocIdPusher, but it seems to be used only in implemented UnsupportedDocIdPusher).

test/com/google/enterprise/adaptor/DocIdSenderTest.java
```diff
@@ -35,6 +30,10 @@
 import java.util.Map;
 import java.util.TreeMap;
 import java.util.concurrent.atomic.AtomicLong;
+import org.junit.Before;
```
wiarlawd: Why do this in this CL?

wiarlawd (commented):
Partial review (I haven't looked at the tests yet).

John L

src/com/google/enterprise/adaptor/DocIdSender.java
```diff
@@ -292,19 +292,7 @@ public GroupPrincipal pushGroupDefinitions(
       handler = defaultErrorHandler;
     }
     boolean firstBatch = true;
-    // if we are not doing an incremental groups push (that is, we are replacing
-    // all groups from the given data source), we must do it in a single "batch"
-    // because if done in batches and a later batch fails, we would end up with
-    // an incomplete set of groups (which could end up giving a user access to
-    // search results that they are not supposed to have access to).
-    if (!incremental) {
-      log.log(Level.INFO, "About to replace all groups from this data source "
```
wiarlawd: Do we log somewhere else whether it was an incremental or full push?

src/com/google/enterprise/adaptor/GsaFeedFileSender.java
BrettMichaelJohnson: Most of the other methods like these have the xmlString as the last param.  At some point, someone added useCompression at the end, then imysak added incremental after that.

test/com/google/enterprise/adaptor/AccumulatingDocIdPusher.java
BrettMichaelJohnson: PJ seems to be resistant to sharing test utility classes. But at least (unlike util classes) there is a somewhat standard place to put them: testing.

test/com/google/enterprise/adaptor/DocIdSenderTest.java
BrettMichaelJohnson: This was a result of the patch from imysak's CL.

src/com/google/enterprise/adaptor/DocIdPusher.java
BrettMichaelJohnson: If it weren't for the fact that ACL also has caseSensitive, that might want to be an enum.

src/com/google/enterprise/adaptor/DocIdPusher.java
```diff
@@ -184,6 +184,43 @@ public GroupPrincipal pushGroupDefinitions(
...
+   * @param defs map of group definitions
+   * @param caseSensitive when comparing Principals
+   * @param incremental if {@code true} incremental update is done, otherwise
+   *        full replacement is done.
+   * @param groupSource
+   * @param handler for dealing with errors pushing
+   * @return {@code null} on success, otherwise the first GroupPrincipal to fail
+   * @throws InterruptedException if interrupted and no definitions were sent
+   */
+  public GroupPrincipal pushGroupDefinitions(
```
PapaJoltOmega: Does adding this method to this interface require any adaptor sources to be modified before they compile with library?

PapaJoltOmega (commented):
Thank you

src/com/google/enterprise/adaptor/DocIdPusher.java
BrettMichaelJohnson: Only those that Mock DocIdPusher, but not those that Proxy DocIdPusher.

BrettMichaelJohnson: Strike that proxy bit, since the abstract class calls the new method.

wiarlawd: I don't know what you mean by proxy here. Only the library uses AbstractDocIdPusher. None of the connectors do. Most of them have implementations of DocIdPusher that will need to be updated, but only in the tests. They also mostly have the same (or very similar) implementations for testing, which is why I suggested we could put them into the library, so avoid this problem in the future.

src/com/google/enterprise/adaptor/DocIdPusher.java
wiarlawd: Grr. Neither place should have a case-sensitive argument, that was a mistake. I would still not add a second boolean next to the first.

test/com/google/enterprise/adaptor/DocIdSenderTest.java
```diff
@@ -491,6 +534,56 @@ public void testAclItemToString() {
...
+    tmp.add(new SimpleImmutableEntry<GroupPrincipal, Collection<Principal>>(
+        new GroupPrincipal("g2"), SAMPLE_DATA.get(new GroupPrincipal("g2"))));
+    tmp.add(new SimpleImmutableEntry<GroupPrincipal, Collection<Principal>>(
+        new GroupPrincipal("g3"), SAMPLE_DATA.get(new GroupPrincipal("g3"))));
+    goldenGroups.add(tmp);
+    return goldenGroups;
+  }
+
+  private static List<List<Map.Entry<GroupPrincipal, Collection<Principal>>>>
+      splitPer2GroupsExpectedResult() {
```
wiarlawd: What about Lists.partion(unsplitExpectedResult().get(0), 2), or something of that nature, or building both off just the inner list?

test/com/google/enterprise/adaptor/DocIdSenderTest.java
```diff
@@ -51,6 +50,15 @@
   private ExceptionHandler runtimeExceptionHandler
       = new RuntimeExceptionExceptionHandler();
 
+  private static final Map<GroupPrincipal, Collection<Principal>> SAMPLE_DATA
```
wiarlawd: I don't like having the test data buried where it is. Maybe at least add a comment here about the contents of each of these?

test/com/google/enterprise/adaptor/DocIdSenderTest.java
```diff
@@ -491,6 +534,56 @@ public void testAclItemToString() {
...
+  }
+
+  private static List<List<Map.Entry<GroupPrincipal, Collection<Principal>>>>
+      unsplitExpectedResult() {
+    List<List<Map.Entry<GroupPrincipal, Collection<Principal>>>> goldenGroups
+        = new ArrayList<List<Map.Entry<GroupPrincipal,
+          Collection<Principal>>>>();
+    List<Map.Entry<GroupPrincipal, Collection<Principal>>> tmp
+        = new ArrayList<Map.Entry<GroupPrincipal, Collection<Principal>>>();
+    tmp.add(new SimpleImmutableEntry<GroupPrincipal, Collection<Principal>>(
```
wiarlawd: I get this was ugly before. My two thoughts to make this more readable, which may not be necessary, because the contents are pretty straightfoward, are to add a newEntry helper method to infer the generic types, or to do something like:

new ArrayList<ugly type>(ImmutableMap.of(...).entrySet())

ImmutableMap uses insertion order for iteration.

wiarlawd (changes requested):
Another partial review.

John L

PapaJoltOmega:
I think developers should not have to update their code to move from v4.1.2 to v4.1.3 . Given that interface is changing that seems like deal breaker for v4.1.3 .

BrettMichaelJohnson:
How would developer's be required to update their code if a library object adds a new method that they don't use?  We will update Documentum, OpenText, and Sharepoint connectors to use the new method, not because it would be required to compile or run, but because it fixes a P2 bug in the connectors.

PapaJoltOmega:
I must have misunderstood one of the comments.  I thought that an interface was being modified that required code change in adaptor instances. I'll take closer look at code.  Thank you

BrettMichaelJohnson:
The comment you are referring to would be our tests that implement a MockDocIdPusher or copies of the library's AccumulatingDocIdPusher.

PapaJoltOmega:
Added reviewer. This reviewer has attempted coding this feature previously. Thank you

src/com/google/enterprise/adaptor/DocIdSender.java
```diff
@@ -211,16 +216,64 @@ public DocId pushNamedResources(Map<DocId, Acl> resources,
...
+   * detect when a group has been deleted, so repeated incremental updates
+   * would leave deleted groups in the GSA's groups database.
+   *
+   * When pushing a full feed, we must do it in a single "batch", because if
+   * done in batches and a later batch fails (or it simply takes a non-trivial
+   * amount of time to upload all the batches), we would end up with an
+   * incomplete set of groups.
+   *
+   * The problem with supplying all the group definitions in a single full
+   * feed is the GSA limit to the size of a feed file (1GB). Although this
```
imysak: I saw such information in the documentation, but I didn't found any evidence of this in the code? Did you try to upload a big feed file e.g. 2-4 GB ?

src/com/google/enterprise/adaptor/DocIdSender.java
```diff
@@ -303,7 +375,9 @@ public GroupPrincipal pushGroupDefinitions(
       log.log(Level.INFO, "Pushing batch of {0} groups", batch.size());
       GroupPrincipal failedId;
       try {
-        failedId = pushSizedBatchOfGroups(batch, caseSensitive, handler);
+        // all groups requests except first should be incremental!
```
imysak: I guess this comment is not correct anymore

src/com/google/enterprise/adaptor/DocIdSender.java
```diff
@@ -337,16 +411,28 @@ public GroupPrincipal pushGroupDefinitions(
...
+    // Remember the latest target group source fed. If we just did a
+    // full group feed, delete the previous target group source entries.
+    String oldGroupSource = groupSources.put(groupSource, feedSourceName);
+    if (feedType == FULL && (oldGroupSource != null)) {
+      // Delete any entries from the older group source by pushing an
+      // empty full feed.
+      log.log(Level.FINE, "Deleting entries from previous group source {0}",
+          oldGroupSource);
+      List<Map.Entry<GroupPrincipal, T>> emptyList = Collections.emptyList();
+      pushSizedBatchOfGroups(emptyList, caseSensitive, FULL, oldGroupSource,
```
imysak, line 431: If I understand updated logic correctly we will have three feed files.
E.g. We have groupSource = "feed". Most Adaptors (as well as AD) have IncrementalListing and FullListing processes.
For IncrementalListing we will use incremental pushes and we will works with data source "feed". 
With FullListing we will use full pushes so we will works with two data sources "feed-A" and "feed-B". 
At the and of "full" push we will clean elder data source from two ("feed-A" or "feed-B"). But in case if some groups was added with IncrementalListing process and was removed it still will be existed in data source "feed". So these groups will never be deleted.
And the same if you launch updated adaptor without resetting index. GSA fill already contains "feed" data source that could contains groups that never be deleted.

It my thoughts are correct we need to fix this. But in fact the fix is elementary. just need to remove "feed" data source as well. I guess:
pushSizedBatchOfGroups(emptyList, caseSensitive, FULL, groupSource, handler);

src/com/google/enterprise/adaptor/DocIdPusher.java
imysak: Adding this method to interface requires modification code(tests but still) for some adaptors (and thirdparty's as well). This doesn't looks like a minor changes and this was the main reason why my CL was rejected. In fact I am agree with this.

Last option that we discussed was the adding public method in abstract class and DocIdSender but not add it in interface. Yes, this require from AD connector to check and cast type to deal with new method but help us to avoid changing code in any adaptors.

Could you review this option?

BrettMichaelJohnson: This method will be used by all (or nearly all) our connectors that push groups, not just AD, so all of their mocks will also need to be updated.  I would consider penetrating the interface to use an undocumented internal method to be poor programming practice.

src/com/google/enterprise/adaptor/DocIdSender.java
BrettMichaelJohnson: One of our largest customers has 100,000 groups, averaging 10,000 members each.  That would result in a 100GB feed.

src/com/google/enterprise/adaptor/DocIdSender.java
BrettMichaelJohnson: I have removed it.

src/com/google/enterprise/adaptor/DocIdSender.java
BrettMichaelJohnson: You are correct. INCREMENTAL feeds should use the name of the last FULL feed pushed, not a different name.

test/com/google/enterprise/adaptor/DocIdSenderTest.java
BrettMichaelJohnson: I left these as  imysak had refactored the original.
I could make a pass through and clean them up.

src/com/google/enterprise/adaptor/DocIdPusher.java
wiarlawd: The original change removed an existing method from the interface, which broke binary and source compatibility with production code. This change breaks source compatibility with unit tests, which is a much smaller change.

The suggested fix is to put the classes that would need modifying in the connectors into the library where they can be updated as needed in the future. (This was proposed and rejected long ago.) Making DocIdPusher an abstract class instead of an interface would also solve the problem in the future. Using a Java 8 default method would be source and binary compatible, though both solutions are a little odd.

The solution through reflection is a little more complicated than just instanceof and a cast, to accommodate the mocks that cause the source incompatibility in the first place. We would need to use reflection to look for and invoke the new method. Making an ugly change in the production code of most connectors seems clearly worse than making a rote change in the unit tests of most connectors.

src/com/google/enterprise/adaptor/DocIdSender.java
```diff
@@ -211,16 +216,78 @@ public DocId pushNamedResources(Map<DocId, Acl> resources,
...
                  + "markAllDocsAsPublic is true.");
       return null;
     }
-    return pushGroupDefinitionsInternal(defs, caseSensitive, handler);
+    return pushGroupDefinitionsInternal(
+        defs, caseSensitive, feedType, groupSource, handler);
+  }
+
+  /*
+   * Issue: Full vs. Incremental Feeds for Groups.
```
wiarlawd: This is really, really verbose. We should document the distinction between FULL and INCREMENTAL on the FeedType enum. Then here, just something like feed files are limited to 1 GB, so we cannot generally push a full set of group definitions in one feed. The solution used here ...

src/com/google/enterprise/adaptor/DocIdSender.java
```diff
@@ -211,16 +216,78 @@ public DocId pushNamedResources(Map<DocId, Acl> resources,
...
+   * definitions for the newer data source.
+   *
+   * If there is a failure mid-push, the GSA groups database may have entries
+   * for both data sources. Since the adaptors do not maintain persistent
+   * state, on a restart it will not know which is "older" vs. "newer".
+   * This is not fatal however, as the only downside might be some deleted
+   * groups persisting through one or two "full" pushes before they get
+   * cleaned up.
+   */
+  private Map<String, String> groupSources = new HashMap<String, String>();
```
wiarlawd: Despite all the verbosity, I don't think you actually said what this is a map of.

src/com/google/enterprise/adaptor/DocIdSender.java
```diff
@@ -337,16 +424,32 @@ public GroupPrincipal pushGroupDefinitions(
       double mean = ((double) numMembers) / numGroups;
       log.finer("mean size of groups: " + mean);
     }
+
+    // Remember the latest target group source fed. If we just did a
+    // full group feed, delete the previous target group source entries.
+    String oldGroupSource;
```
wiarlawd: Can we just move this down into the if statement? If we never do a full feed, why not just avoid ever writing to the groupSources map?

src/com/google/enterprise/adaptor/DocIdPusher.java
```diff
@@ -184,6 +200,43 @@ public GroupPrincipal pushGroupDefinitions(
...
+   * If groupSource is {@code null}, then the {@code feed.name} configuration
+   * property is used.
+   *
+   * <p>If handler is {@code null}, then a default error handler is used.
+   *
+   * @param defs map of group definitions
+   * @param caseSensitive when comparing Principals
+   * @param feedType if INCREMENTAL, an incremental update is done; if FULL, a
+   *        full replacement is done.
+   * @param groupSource
```
wiarlawd, line 257: We're stealing 2 of the 256 allowed characters for the data source name, so we should mention that the max length here is 254.

src/com/google/enterprise/adaptor/DocIdSender.java
```diff
@@ -211,16 +216,78 @@ public DocId pushNamedResources(Map<DocId, Acl> resources,
...
+   * eventually get deleted from the GSA.
+   */
+  private String previousGroupSource(String source) {
+    String prevSource = groupSources.get(source);
+    return (prevSource == null) ? source : prevSource;
+  }
+
+  /** Returns the alternate to the previous group source. */
+  private String alternateGroupSource(String source) {
+    return (source + "-A").equals(groupSources.get(source))
```
wiarlawd: These names are admin-visible in the admin console. Is there something more intuitive? I think numbers might be slightly more explicable, in that splitting foo into foo-1 and foo-2 might be more idiomatic than foo-A and foo-B. Maybe even -full-1 and -full-2? The datasource name length is 256, so we're not stealing characters from a particularly rare resource (I thought the limit was 32, for example).

For posterity, I didn't like the use of 3 possible name sources, but I came to like the fact that the use of the decorated names indicates that a full feed was used at some point. Might be useful info for debugging.

src/com/google/enterprise/adaptor/DocIdSender.java
```diff
@@ -273,15 +340,34 @@ public GroupPrincipal pushGroupDefinitions(
...
+      groupSource = config.getFeedName();
+    }
+    log.log(Level.INFO, "Starting {0} push of {0} groups from source {1}",
+        new Object[] { feedType, defs.size(), groupSource });
+    String feedSourceName;
+    if (feedType == INCREMENTAL) {
+      feedSourceName = previousGroupSource(groupSource);
+    } else {
+      feedSourceName = alternateGroupSource(groupSource);
+      log.log(Level.FINE, "Using alternate group source {0}", feedSourceName);
```
wiarlawd: Maybe s/Using/Writing to/?

src/com/google/enterprise/adaptor/DocIdSender.java
```diff
@@ -355,8 +458,8 @@ public GroupPrincipal pushGroupDefinitions(
     for (int ntries = 1; keepGoing; ntries++) {
       try {
         log.info("sending groups to GSA host name: " + config.getGsaHostname());
-        fileSender.sendGroups(feedSourceName,
-            groupsDefXml, config.isServerToUseCompression());
+        fileSender.sendGroups(feedSourceName, feedType.toString(), groupsDefXml,
```
wiarlawd: It's me, the git history pedant! Why modify the second line if you don't have to?

test/com/google/enterprise/adaptor/DocIdSenderTest.java
```diff
@@ -14,16 +14,16 @@
...
 package com.google.enterprise.adaptor;
 
+import static com.google.enterprise.adaptor.DocIdPusher.EVERYTHING_CASE_SENSITIVE;
+import static com.google.enterprise.adaptor.DocIdPusher.FeedType.FULL;
+import static com.google.enterprise.adaptor.DocIdPusher.FeedType.INCREMENTAL;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertTrue;
 
+import com.google.common.collect.ImmutableList;
```
wiarlawd, line 24: Avoid Guava. This class was explicitly using Java Collections instead.

src/com/google/enterprise/adaptor/DocIdSender.java
```diff
@@ -337,16 +424,32 @@ public GroupPrincipal pushGroupDefinitions(
       double mean = ((double) numMembers) / numGroups;
       log.finer("mean size of groups: " + mean);
     }
+
+    // Remember the latest target group source fed. If we just did a
+    // full group feed, delete the previous target group source entries.
+    String oldGroupSource;
+    synchronized(groupSources) {
```
wiarlawd: Add space after "synchronized".

test/com/google/enterprise/adaptor/DocIdSenderTest.java
wiarlawd: Can we undo this? It generates a lint error using the old linter, and no one has gotten the new linter working yet.

test/com/google/enterprise/adaptor/DocIdSenderTest.java
```diff
@@ -491,6 +564,48 @@ public void testAclItemToString() {
...
+
+  private static Map<GroupPrincipal, Collection<Principal>> emptyGroups() {
+    return ImmutableMap.<GroupPrincipal, Collection<Principal>>of();
+  }
+
+  // TODO(bmj): Uncomment @SafeVarargs when moving to Java 7 and remove the
+  // @SuppressWarnings("unchecked") usages above.
+  // @SafeVarargs
+  private static List<List<Map.Entry<GroupPrincipal, Collection<Principal>>>>
+      expectedResult(int batchSize,
```
wiarlawd: Could you document what this does? And write a unit test for it? I presume it takes the group definitions as inputs and generates the expected batches recorded by the mock?

test/com/google/enterprise/adaptor/DocIdSenderTest.java
```diff
@@ -491,6 +564,48 @@ public void testAclItemToString() {
     assertEquals(golden, "" + new DocIdSender.AclItem(id, acl));
   }
 
+  private static Map<GroupPrincipal, Collection<Principal>> sampleGroups() {
```
wiarlawd: Please move this back to the top. I dislike buried test data, it makes the tests read like too much magic. I assume you created this method because having the test data replicated in each test was too messy? I'm not sure that I entirely agree, but at least let's put this back first and see how it reads.

test/com/google/enterprise/adaptor/GsaFeedFileSenderTest.java
```diff
@@ -36,8 +29,12 @@
 import java.net.URL;
 import java.nio.charset.Charset;
 import java.util.zip.GZIPInputStream;
-
 import javax.net.ssl.SSLContext;
+import org.junit.After;
```
wiarlawd: Undo

test/com/google/enterprise/adaptor/DocIdSenderTest.java
```diff
@@ -348,73 +352,142 @@ public void getModifiedDocIds(DocIdPusher pusher) throws IOException, Interrupte
...
+
+    assertNull(docIdSender.pushGroupDefinitions(sampleGroups(),
+        EVERYTHING_CASE_SENSITIVE, FULL, null, null));
+
+    assertEquals(Collections.singletonList("default_source"),
+        fileSender.groupsources);
+  }
+
+  @Test
+  public void testPushGroupsGroupSource() throws Exception {
```
wiarlawd, line 636: I feel like this quit right at is got interesting. Why aren't you checking the group source names in the tests above where you are expecting overflow to the alternate? I would also think that we want tests that roll from foo to foo-A and delete foo, and that roll to foo-B and delete foo-A, and make sure that incremental feeds along the way go into foo, foo-A, and foo-B as expected.

wiarlawd (changes requested):

src/com/google/enterprise/adaptor/DocIdPusher.java
BrettMichaelJohnson: Done.

src/com/google/enterprise/adaptor/DocIdSender.java
BrettMichaelJohnson: The 1GB limit is explicitly configured in the GSA's code.

src/com/google/enterprise/adaptor/DocIdSender.java
BrettMichaelJohnson: DOne.

src/com/google/enterprise/adaptor/DocIdSender.java
BrettMichaelJohnson: I like the FULL1 and FULL2 designations.

src/com/google/enterprise/adaptor/DocIdSender.java
BrettMichaelJohnson: Done.

src/com/google/enterprise/adaptor/DocIdSender.java
BrettMichaelJohnson: Done.

src/com/google/enterprise/adaptor/DocIdSender.java
BrettMichaelJohnson: Done.

src/com/google/enterprise/adaptor/DocIdSender.java
BrettMichaelJohnson: Done.

test/com/google/enterprise/adaptor/DocIdSenderTest.java
BrettMichaelJohnson: Undone.

test/com/google/enterprise/adaptor/GsaFeedFileSenderTest.java
BrettMichaelJohnson: Undone.

test/com/google/enterprise/adaptor/DocIdSenderTest.java
BrettMichaelJohnson: Done.

test/com/google/enterprise/adaptor/DocIdSenderTest.java
BrettMichaelJohnson: Done.

src/com/google/enterprise/adaptor/DocIdPusher.java
```diff
@@ -31,7 +31,18 @@
...
+   * FeedType used for group definition pushes.
+   *
+   * GSA 7.4.0 added support for full feeds of group definitions. A {@code FULL}
+   * feed replaces all groups from a given data source. An {@code INCREMENTAL}
+   * feed augments or updates existing group definitions for a given data source.
+   * When feeding GSAs earlier than 7.4.0, {@code INCREMENTAL} feeds are always
+   * used. {@code INCREMENTAL} is the default FeedType for the overloaded
+   * {@code pushGroupDefinitions} methods that do specify a FeedType.
+   *
+   * @since 4.1.4
```
wiarlawd, line 44: I think @since TODO(bmj) would be better. We had a habit with v3 of looking for "@since TODO" before each release. You could keep this and add the other in an implementation comment just after this. /* @since TODO(bmj) */

src/com/google/enterprise/adaptor/DocIdPusher.java
```diff
@@ -215,8 +226,9 @@ public GroupPrincipal pushGroupDefinitions(
    * will wholly replace all existing definitions for the groupSource on the
    * GSA.
    *
-   * <p>A groupSource identifies the source of the group definitions. If
-   * provided, it must be a string of the form {@code [a-zA-Z_][a-zA-Z0-9_-]*}.
+   * <p>A groupSource identifies the source of the group definitions.
+   * If provided, it must be a string of the form
+   * <code>[a-zA-Z_][a-zA-Z0-9_-]{0,250}</code>.
```
wiarlawd: I think this is too technical, and you should stick to the * and give the max length separately.

src/com/google/enterprise/adaptor/DocIdSender.java
```diff
@@ -268,6 +257,9 @@ public GroupPrincipal pushGroupDefinitions(
    * This is not fatal however, as the only downside might be some deleted
    * groups persisting through one or two "full" pushes before they get
    * cleaned up.
+   *
+   * This maps the connector-supplied to the last used pseudonym, e.g.
```
wiarlawd: I think the is missing from your sentence.

test/com/google/enterprise/adaptor/DocIdSenderTest.java
```diff
@@ -437,36 +584,33 @@ public void testPushGroupsReplaceAllGroupsSingleFeed() throws Exception {
...
         expectedResult(Integer.MAX_VALUE, sampleGroups(), emptyGroups()),
         fileMaker.groupses);
     assertEquals(ImmutableList.of("incremental", "full"), fileSender.feedtypes);
+    assertEquals(ImmutableList.of("foo-FULL1", "foo"), fileSender.groupsources);
 
     assertNull(docIdSender.pushGroupDefinitions(sampleGroups(),
-        EVERYTHING_CASE_SENSITIVE, FULL, null, null));
+        EVERYTHING_CASE_SENSITIVE, FULL, "foo", null));
 
     assertEquals(4, fileMaker.i);
```
wiarlawd: Maybe add a comment before this group that the data collected by the mock fileSender is cumulative and includes previous results.

test/com/google/enterprise/adaptor/DocIdSenderTest.java
```diff
@@ -347,74 +353,282 @@ public void getModifiedDocIds(DocIdPusher pusher) throws IOException, Interrupte
...
+    return ImmutableMap.<GroupPrincipal, Collection<Principal>>of();
+  }
+
+  /**
+   * Returns the group definitions broken into batches of a batchSize entries.
+   */
+  // TODO(bmj): Uncomment @SafeVarargs when moving to Java 7 and remove the
+  // @SuppressWarnings("unchecked") usages below.
+  // @SafeVarargs
+  private static List<List<Map.Entry<GroupPrincipal, Collection<Principal>>>>
```
wiarlawd: I didn't mean to suggest moving this up. I would keep this and its tests buried.

wiarlawd (changes requested):

test/com/google/enterprise/adaptor/DocIdSenderTest.java
```diff
@@ -347,74 +353,282 @@ public void getModifiedDocIds(DocIdPusher pusher) throws IOException, Interrupte
...
+                    new UserPrincipal("u1")),
+                new GroupPrincipal("g2"), ImmutableList.<Principal>of(
+                    new UserPrincipal("u2")),
+                new GroupPrincipal("g3"), ImmutableList.<Principal>of(
+                    new UserPrincipal("u3"))),
+            ImmutableMap.<GroupPrincipal, Collection<Principal>>of(
+                new GroupPrincipal("g4"), ImmutableList.<Principal>of(
+                    new UserPrincipal("u4")),
+                new GroupPrincipal("g5"), ImmutableList.<Principal>of(
+                    new UserPrincipal("u5"))));
```
wiarlawd: Missing )

test/com/google/enterprise/adaptor/DocIdSenderTest.java
```diff
@@ -347,74 +353,282 @@ public void getModifiedDocIds(DocIdPusher pusher) throws IOException, Interrupte
...
+
+  @Test
+  @SuppressWarnings("unchecked")
+  public void testExpectedResultMultipleGroupMaps() {
+    assertEquals(
+        ImmutableList.<List>of(
+            ImmutableList.<Map.Entry>of(
+                mapEntry("g1", "u1"),
+                mapEntry("g2", "u2")),
+            ImmmutableList.<Map.Entry>of(
```
wiarlawd: s/Immmutable/Immutable/, plus another above.

wiarlawd (changes requested):
Compile errors.

src/com/google/enterprise/adaptor/DocIdPusher.java
BrettMichaelJohnson: This should probably move to DocIdPusher2

PapaJoltOmega (commented):
Are the next steps:
(1) Put new method in sub interface
(2) Send for review by Igor

wiarlawd:
Igor's already a reviewer, and the subinterface changes are in #45. I'm not a fan overall of the subinterface, I think it's needless complexity for a very narrow problem. We added a method to Response in 4.1.0 without fanfare, and we are adding machinery to make it an easy problem that you can fix once and not have to touch again (again, in #45 and #46 and others to follow once the pattern is approved).

PapaJoltOmega:
I'm not following how a different CL having a subinterface for a different reason makes this CL progress.

I'm also not following this statement "we are adding machinery to make it an easy problem that you can fix once and not have to touch again".  Please take me throught it in smaller steps. Thank you

wiarlawd:
Sorry, I was rushing out the door. I sent a longer email discussing the broader collection of pull requests adding methods to interfaces, the use of subinterfaces, and publishing test doubles to avoid similar problems in the future.

Here I'll just add that I made the compatibility changes in separate pull requests because this one was already messy, and it allowed us to ponder the compatibility options without the clutter of functional changes.

src/com/google/enterprise/adaptor/DocIdPusher.java
```diff
@@ -43,6 +43,7 @@
    *
    * @since 4.1.4
    */
+  // @since TODO(bmj)
```
wiarlawd: Belay that order.

src/com/google/enterprise/adaptor/DocIdPusher.java
```diff
@@ -245,6 +246,7 @@ public GroupPrincipal pushGroupDefinitions(
...
```
wiarlawd: Belay that order.

src/com/google/enterprise/adaptor/DocIdPusher.java
wiarlawd: bump

src/com/google/enterprise/adaptor/DocIdSender.java
wiarlawd: bump

test/com/google/enterprise/adaptor/DocIdSenderTest.java
wiarlawd: bump

wiarlawd (changes requested):

PapaJoltOmega:
Igor, please read this CL again. This time please read with the idea that we can change the interface and force developers using library to change their code. Thank you

src/com/google/enterprise/adaptor/DocIdPusher.java
```diff
@@ -184,6 +211,46 @@ public GroupPrincipal pushGroupDefinitions(
...
+   * a predictable iteration order, like {@link java.util.TreeMap}.
+   *
+   * <p>An incremental push augments or updates any existing group definitions
+   * for the groupSource on the GSA. Otherwise, the supplied group definitions
+   * will wholly replace all existing definitions for the groupSource on the
+   * GSA.
+   *
+   * <p>A groupSource identifies the source of the group definitions.
+   * If provided, it must be a string of the form
+   * <code>[a-zA-Z_][a-zA-Z0-9_-]{0,250}</code>.
```
wiarlawd: I think this is too technical, and should go back to * with the 250 character limit mentioned separately.

src/com/google/enterprise/adaptor/DocIdSender.java
```diff
@@ -211,16 +216,70 @@ public DocId pushNamedResources(Map<DocId, Acl> resources,
...
+   * definitions for the newer data source.
+   *
+   * If there is a failure mid-push, the GSA groups database may have entries
+   * for both data sources. Since the adaptors do not maintain persistent
+   * state, on a restart it will not know which is "older" vs. "newer".
+   * This is not fatal however, as the only downside might be some deleted
+   * groups persisting through one or two "full" pushes before they get
+   * cleaned up.
+   *
+   * This maps the connector-supplied to the last used pseudonym, e.g.
```
wiarlawd: You're missing a word after "connector-supplied".

wiarlawd (approved):

imysak:
PJ, I reviewed this CL and I have no objections (with the idea that we can change the interface).
LGTM

src/com/google/enterprise/adaptor/DocIdSender.java
```diff
@@ -273,15 +332,35 @@ public GroupPrincipal pushGroupDefinitions(
...
+          gsaVerString);
+      feedType = INCREMENTAL;
+    }
     if (null == handler) {
       handler = defaultErrorHandler;
     }
+    if (groupSource == null) {
+      groupSource = config.getFeedName();
+    }
+    log.log(Level.INFO, "Starting {0} push of {0} groups from source {1}",
```
PapaJoltOmega: should that be {0} {1} {2}

src/com/google/enterprise/adaptor/AbstractDocIdPusher.java
```diff
@@ -66,11 +68,27 @@ public DocId pushNamedResources(Map<DocId, Acl> resources)
...
-    return pushGroupDefinitions(defs, caseSensitive, null);
+    return pushGroupDefinitions(defs, caseSensitive, INCREMENTAL, null, null);
+  }
+
+  /**
+   * Calls {@code pushGroupDefinitions(defs, caseSensitive, INCREMENTAL, null,
+   * handler)}.
+   */
+  @Override
+  public GroupPrincipal pushGroupDefinitions(
```
PapaJoltOmega: Reason to add this method impl is that it should have already been here?

src/com/google/enterprise/adaptor/DocIdPusher.java
```diff
@@ -184,6 +211,46 @@ public GroupPrincipal pushGroupDefinitions(
...
+   * @param feedType if INCREMENTAL, an incremental update is done; if FULL, a
+   *        full replacement is done.
+   * @param groupSource
+   * @param handler for dealing with errors pushing
+   * @return {@code null} on success, otherwise the first GroupPrincipal to fail
+   * @throws InterruptedException if interrupted and no definitions were sent
+   *
+   * @since 4.1.4
+   */
+  public GroupPrincipal pushGroupDefinitions(
```
PapaJoltOmega, line 264: should we change javadoc of other pushGroupDefinitions to say that they are incremental?

src/com/google/enterprise/adaptor/DocIdPusher.java
```diff
@@ -184,6 +211,46 @@ public GroupPrincipal pushGroupDefinitions(
...
+   * If groupSource is {@code null}, then the {@code feed.name} configuration
+   * property is used.
+   *
+   * <p>If handler is {@code null}, then a default error handler is used.
+   *
+   * @param defs map of group definitions
+   * @param caseSensitive when comparing Principals
+   * @param feedType if INCREMENTAL, an incremental update is done; if FULL, a
+   *        full replacement is done.
+   * @param groupSource
```
PapaJoltOmega, line 257: I'm surprised that we need to allow the client to control groupSource. 

Why does client need to control this name?

src/com/google/enterprise/adaptor/DocIdSender.java
```diff
@@ -211,16 +216,70 @@ public DocId pushNamedResources(Map<DocId, Acl> resources,
...
+   */
+  private Map<String, String> groupSources = new HashMap<String, String>();
+
+  /**
+   * Returns the previous group source pushed to. If there was no previous
+   * group source, returns the supplied source. If a connector only ever does
+   * INCREMENTAL feeds, then that group source will always be used. This is
+   * also backward compatible. If a connector is upgraded to a new version that
+   * supports FULL feeds, then the older incremental group source will
+   * eventually get deleted from the GSA.
```
PapaJoltOmega: how will it get deleted?

src/com/google/enterprise/adaptor/DocIdSender.java
```diff
@@ -211,16 +216,70 @@ public DocId pushNamedResources(Map<DocId, Acl> resources,
...
+   *
+   * This maps the connector-supplied group source to the last used pseudonym,
+   * e.g. {@code foo -> foo-FULL1}.
+   */
+  private Map<String, String> groupSources = new HashMap<String, String>();
+
+  /**
+   * Returns the previous group source pushed to. If there was no previous
+   * group source, returns the supplied source. If a connector only ever does
+   * INCREMENTAL feeds, then that group source will always be used. This is
```
PapaJoltOmega, line 265: that group source meaning the supplied source?

src/com/google/enterprise/adaptor/DocIdSender.java
```diff
@@ -211,16 +216,70 @@ public DocId pushNamedResources(Map<DocId, Acl> resources,
...
+   * eventually get deleted from the GSA.
+   */
+  private String previousGroupSource(String source) {
+    String prevSource = groupSources.get(source);
+    return (prevSource == null) ? source : prevSource;
+  }
+
+  /** Returns the alternate to the previous group source. */
+  private String alternateGroupSource(String source) {
+    return (source + "-FULL1").equals(groupSources.get(source))
```
PapaJoltOmega, line 279: groupSources are only added for full?

src/com/google/enterprise/adaptor/DocIdSender.java
```diff
@@ -273,15 +332,35 @@ public GroupPrincipal pushGroupDefinitions(
...
+    if (groupSource == null) {
+      groupSource = config.getFeedName();
+    }
+    log.log(Level.INFO, "Starting {0} push of {0} groups from source {1}",
+        new Object[] { feedType, defs.size(), groupSource });
+    String feedSourceName;
+    if (feedType == INCREMENTAL) {
+      feedSourceName = previousGroupSource(groupSource);
+    } else {
+      feedSourceName = alternateGroupSource(groupSource);
```
PapaJoltOmega, line 358: One important observation is that for purposes of ACL resolution on GSA the group source is not relevant, yes?  That is, group source is relevant for managing the group source's data and potentially clearing it up wholly, but from run to run the group names and group members are the same no matter what name was used for group-source. Is that right?

src/com/google/enterprise/adaptor/DocIdSender.java
```diff
@@ -303,7 +382,8 @@ public GroupPrincipal pushGroupDefinitions(
       log.log(Level.INFO, "Pushing batch of {0} groups", batch.size());
       GroupPrincipal failedId;
       try {
-        failedId = pushSizedBatchOfGroups(batch, caseSensitive, handler);
+        failedId = pushSizedBatchOfGroups(batch, caseSensitive,
```
PapaJoltOmega, line 383: if in FULL mode, we do not have to clear the alternate source?

PapaJoltOmega (commented):
Thank you

src/com/google/enterprise/adaptor/AbstractDocIdPusher.java
BrettMichaelJohnson: No. I wanted to keep only one overloaded pushGroupDefinitions in DocIdSender.  Abstract DocIdPusher has overloads with fewer params that delegate to the one with more params. This 3-argument version of pushGroupDefs was the one in DocIdSender.  Since DocIdSender now as a 5-argument pushGroupDefs, the 3-arg one gets demoted to here and calls the 5 arg version.

src/com/google/enterprise/adaptor/DocIdPusher.java
BrettMichaelJohnson: This will allow the SharePoint connector to push a separate full group feed for each SharePoint site collection, since each has its own distinct groups.

src/com/google/enterprise/adaptor/DocIdPusher.java
BrettMichaelJohnson: Done.

src/com/google/enterprise/adaptor/DocIdSender.java
BrettMichaelJohnson: This method returns the actual source name used in the previous feed. e.g. foo-FULL1.

src/com/google/enterprise/adaptor/DocIdSender.java
BrettMichaelJohnson: Though the magic of the second sentence in this javadoc comment: "If there was no previous group source, returns the supplied source", so the group source name used by older versions of the connector will be the "previous" group source and get deleted once the full push finishes.  However, this will only work when the connector uses the default group source name (feed.name), which was all that was used in previous versions of the library.

src/com/google/enterprise/adaptor/DocIdSender.java
BrettMichaelJohnson: Group sources are used for both full and incremental pushes. Alternating group sources are only used for full pushes as discussed in the "Issue: Full vs. Incremental Feeds for Groups." above.

src/com/google/enterprise/adaptor/DocIdSender.java
BrettMichaelJohnson: Yes it should.

src/com/google/enterprise/adaptor/DocIdSender.java
BrettMichaelJohnson: I would think that ACL resolution on the GSA would be the same as if you had multiple connector instances, each feeding groups using a different feed.name.

src/com/google/enterprise/adaptor/DocIdSender.java
BrettMichaelJohnson: That happens if the full push succeeds, on lines 430-433, below.

src/com/google/enterprise/adaptor/AbstractDocIdPusher.java
PapaJoltOmega: OK. Makes sense.  If you were to redo the interface from scratch, would the methods you'd have be different? (we're already breaking interface definition, so should we go further?)

src/com/google/enterprise/adaptor/DocIdPusher.java
```diff
@@ -24,6 +26,31 @@
  * Interface that allows at-will pushing of {@code DocId}s to the GSA.
  */
 public interface DocIdPusher {
+
+  /** Not as good as an enum for case sensitivity, but backward compatible. */
```
PapaJoltOmega, line 30: How painful would it be to break backwards compatibility?  (we're already breaking the interface, so would further change be a large incremental pain?)

src/com/google/enterprise/adaptor/DocIdPusher.java
PapaJoltOmega: Does the addition of FULL to INCREMENTAL require multiple group names? Should those be seperate changes or do they belong together conceptually?

src/com/google/enterprise/adaptor/DocIdSender.java
```diff
@@ -211,16 +216,70 @@ public DocId pushNamedResources(Map<DocId, Acl> resources,
...
                  + "markAllDocsAsPublic is true.");
       return null;
     }
-    return pushGroupDefinitionsInternal(defs, caseSensitive, handler);
+    return pushGroupDefinitionsInternal(
+        defs, caseSensitive, feedType, groupSource, handler);
+  }
+
+  /**
+   * Issue: Full vs. Incremental Feeds for Groups.
```
PapaJoltOmega: Would it be more accurate to say the issue is "When full group pushes  exceed permissible size of single feed".

The issue's descroption seems right. The label of the issue seems not right.

src/com/google/enterprise/adaptor/DocIdSender.java
```diff
@@ -211,16 +216,70 @@ public DocId pushNamedResources(Map<DocId, Acl> resources,
...
+   * feed is the GSA limit to the size of a feed file (1GB). Although this
+   * is a large limit, several of our customers would easily exceed that.
+   * One customer has 100,000 groups averaging 10,000 members each. Another
+   * has 400,000 groups.
+   *
+   * A novel solution to this is to double-buffer the group definitions for
+   * each data source. We maintain two pseudo data sources for each actual
+   * data source (*-FULL1 and *-FULL2). Each "full" upload alternates between
+   * these two, using batches of incremental feeds. As each batch is processed,
+   * groups defined in that batch are re-assigned from the older data source
```
PapaJoltOmega: re-assigned seems unusual to me. 
Maybe consider
s/re-assigned/replace groups/
or maybe
s/re-assigned/override/

src/com/google/enterprise/adaptor/DocIdSender.java
```diff
@@ -211,16 +216,70 @@ public DocId pushNamedResources(Map<DocId, Acl> resources,
...
+   * groups persisting through one or two "full" pushes before they get
+   * cleaned up.
+   *
+   * This maps the connector-supplied group source to the last used pseudonym,
+   * e.g. {@code foo -> foo-FULL1}.
+   */
+  private Map<String, String> groupSources = new HashMap<String, String>();
+
+  /**
+   * Returns the previous group source pushed to. If there was no previous
```
PapaJoltOmega, line 263: Is this sentence true for both incremental and full?

src/com/google/enterprise/adaptor/DocIdSender.java
PapaJoltOmega: With current imple if client does:
1. incremental feeds to "foo"
2. full push to "foo" (which is full push to "foo-FULL1")
then "foo" gets deleted.

That seems like what Igor was suggesting. I don't follow the comment then that incremental should use the name of last FULL feed pushed.  Does that mean, that if client (3) pushes more incremental to "foo" those foo will go to "foo-FULL1"?

test/com/google/enterprise/adaptor/DocIdSenderTest.java
```diff
@@ -347,74 +353,282 @@ public void getModifiedDocIds(DocIdPusher pusher) throws IOException, Interrupte
...
+    assertEquals(ImmutableList.of("incremental", "full"), fileSender.feedtypes);
+    assertEquals(ImmutableList.of("foo-FULL1", "foo"), fileSender.groupsources);
+    assertEquals(ImmutableList.of("0", "1"), fileSender.xmlStrings);
+    assertEquals(ImmutableList.of("0", "1"), fileArchiver.feeds);
+    assertTrue(fileArchiver.failedFeeds.isEmpty());
+  }
+
+  @Test
+  @SuppressWarnings("unchecked")
+  public void testPushGroupsReplaceAllGroupsAlternateBuffer() throws Exception {
```
PapaJoltOmega, line 454: do we have tests for:
(1) incremental, follow by full,
(2) incremental, follow by full, follow by full
(3) full, incremental, incremental
(4) full, incremental, full
(5) full, incremental, full, incremental

PapaJoltOmega (commented):
Sorry for the extensive delay. Thank you for your patience.

src/com/google/enterprise/adaptor/AbstractDocIdPusher.java
BrettMichaelJohnson: TBH, I don't like this many overloads. All our connectors that push groups will be using the 5-arg pusher, so the rest remain for other connectors in the wild. If I was redesigning from scratch, I would probably be tempted to create a fluent GroupPusher.Builder.

src/com/google/enterprise/adaptor/DocIdPusher.java
BrettMichaelJohnson: I think it would be problematic to break backward compatibility, especially in a +.01 release. The change just makes for more readable code.

src/com/google/enterprise/adaptor/DocIdPusher.java
BrettMichaelJohnson: At this point, only the Sharepoint connector needs multiple group sources, and each of them will be a FULL feed, so it makes sense to pair the features. IIRC, the SP connector also detects modified groups, so those will be pushed as INCREMENTAL changes with the respective group source. The rest of the connectors have a single group source.

src/com/google/enterprise/adaptor/DocIdSender.java
BrettMichaelJohnson: Done.

src/com/google/enterprise/adaptor/DocIdSender.java
BrettMichaelJohnson: My problem is I am unclear as to how the GSA does group resolution, especially if two different group sources have identical group names.  I suspect this sentence and the next are completely inaccurate.

src/com/google/enterprise/adaptor/DocIdSender.java
BrettMichaelJohnson: Yes.

src/com/google/enterprise/adaptor/DocIdSender.java
BrettMichaelJohnson: Precisely. Some connectors support modified group detection, so those modifications would apply to the groups sent in the last full feed, in your case foo-FULL1.

test/com/google/enterprise/adaptor/DocIdSenderTest.java
BrettMichaelJohnson: Added new tests:
1) First 2/3 of testPushGroupsIncrementalThenFullThenFull
2) testPushGroupsIncrementalThenFullThenFull
3) testPushGroupsFullThenIncrementalThenIncremental
4) First 3/4 of testPushGroupsFullThenIncrementalThenFullThenIncremental
5) testPushGroupsFullThenIncrementalThenFullThenIncremental

I also added a test for feeding from multiple group sources.
These additional tests are exercising the previousGroupSource()
and alternateGroupSource() methods for selecting which feed
datasource name to use.

src/com/google/enterprise/adaptor/DocIdSender.java
```diff
@@ -211,16 +216,68 @@ public DocId pushNamedResources(Map<DocId, Acl> resources,
...
+   * group source, returns the supplied source. If a connector only ever does
+   * INCREMENTAL feeds, then that group source will always be used. This is
+   * also backward compatible. If a connector is upgraded to a new version that
+   * supports FULL feeds, then the older incremental group source will
+   * eventually get deleted from the GSA. If a connector does both FULL and
+   * INCREMENTAL feeds, then an incremental feed will the the last FULL feed
```
PapaJoltOmega: s/the the/use the/

PapaJoltOmega (approved):
Thank you

PR #28, opened 2017-04-13
Upgrade to JUnit 4.11

BrettMichaelJohnson:
Not the latest release, but the one used in Cobertura.


wiarlawd:
This is incomplete (build.xml has to change), it also requires a CL for every connector (which makes it bigger than it seems), and I had tabled this for post-4.1.3 because it isn't important enough to change just prior to a release (see b/36821646).

John L

BrettMichaelJohnson:
> This is incomplete (build.xml has to change)

Dang! I was tripped up by the same "This seems to work" as you were.

> I had tabled this for post-4.1.3 because it isn't important enough to change just prior to a release 

Sorry, I was just going through my list of things we talked about.  I agree, this can wait. I will leave the CL up until after the release.

PapaJoltOmega:
Thanks jlacey@ for catching. Developers don't need to change code (or build file) between v4.1.2 and v4.1.3 right now, and I believe that is an important charateristic to maintain. Thank you

build.xml
```diff
@@ -123,6 +122,19 @@
...
+  <path id="junit.build.classpath">
+    <fileset dir="${lib.dir}">
+      <include name="junit-4.11.jar"/>
+    </fileset>
+  </path>
+
+  <path id="junit.run.classpath">
+    <path refid="junit.build.classpath"/>
+    <fileset dir="${lib.dir}">
+      <include name="hamcrest-core-1.3.jar"/>
```
wiarlawd: How do you end up needing this at runtime and not at compile-time?

BrettMichaelJohnson: We use no custom matchers in our tests, although JUnit uses it under the covers. I suppose I could add it to the compile path in case a connector's unit tests want to use custom matchers. Another possibility compilation succeeded might be that it was picking up the ham crest jar from Corbatura.

wiarlawd: It turns out that the org.hamcrest classes needed to load the JUnit classes, e.g., Assert, are in junit.jar itself. The hamcrest-core.jar file contains extra classes that might be referenced, but those cannot be needed at runtime unless they were referenced by our code at compile-time.

So we can either just update junit.jar to preserve the status-quo, or include hamcrest-core.jar at compile-time to allow ourselves more freedom.

I'm mildly inclined to use Truth rather than Hamcrest if we want a more extended framework with better error messages, and to stick with just the raw JUnit for now.

wiarlawd (commented):
LGTM, but please wait for PJ's approval, and we need to wait for the 4.1.3 release before committing this, or changing the connectors.

John L

PapaJoltOmega:
I'm not following how hamcrest makes it into the picture. I don't follow how cobertura makes hamcrest come into the picture. I think i expected what John initially expected:
- add junit jar
- delete previous junit jar
- 1 line change build.xml
- 1 line change THIRDPARTYLICENSE.txt

wiarlawd:
At some point prior our current version, 4.8.2, JUnit embedded the Hamcrest core matchers (an assertion framework much like Truth). In 4.8.2, specifically, the minimally needed Hamcrest classes are actually in junit-4.8.2.jar, and not in the accompanying hamcrest-core-*.jar. For example, there is an assertThat(T, Matcher) method in Assert, which means that referencing the Assert class requires loading Matcher as well.

Brett said that later versions, I think it was staring with 4.9.something, the minimally needed Hamcrest classes were moved back to the hamcrest-core-*.jar. That means that you cannot reference the Assert class at all without having both junit-*.jar and hamcrest-core-*.jar on the class path.

Cobertura was a red herring. Brett thought that maybe his original commit worked because javac was finding the Hamcrest classes in Cobertura, but it was just broken and didn't compile. The hamcrest-core-1.3.jar is required at both compile-time and runtime, and Cobertura is not referenced at compile-time.

Good catch on the THIRDPARTYLICENSE.txt. I had missed that when I updated Guava.

PapaJoltOmega:
OK.  Got it.  So everyone that decides to use junit > 4.9 needs to include the hamcrest jar.  

We're not including a new dep.  It is that one of our deps is now two jars instead of one.

And, per seperate chat, adaptor instance developers can update to library version past this CL here and `ant test' and `ant dist' without making source or build file changes or any other changes.

LGTM

PapaJoltOmega:
I'm thinking the THIRDPARTYLICENSE.txt change for guava can be my practice run of a github pull request.  Thank you.

PapaJoltOmega (approved):

wiarlawd:
FYI, this will conflict with #16, so whichever goes second needs to make some changes before committing. That will probably be you, if we're holding this pull request until 4.1.3 is released.

John L

wiarlawd:
#16 is committed, so this PR could be updated now, including THIRDPARTYLICENSE.txt.

John L

PapaJoltOmega (approved):

wiarlawd (approved):
I approve this for merging to master during the code slush, as it's low risk and affects only the tests.

PR #29, opened 2017-04-13
Add RegexDecisionFilter MetadataTransform

BrettMichaelJohnson:
RegexDecisionFilter makes a TransmissionDecision using regular expression matching on metadata and/or params values.  It is basically SkipDocumentFilter with a configurable decision. RegexDecisionFilter takes the following configuration parameters:

key: the metadata or parameter key whose values will be matched against pattern.

pattern: a regular expression used to match values. if unspecified, anything will match (effectively tests presence/absence of key).

decideOnMatch: if true the decision is applied if a match is found. if false the decision is applied if no match is found.

decision: one of TransmissionDecision string values: "as-is", "do-not-index", "do-not-index-content".

This pull request contains a series of commits that are incremental changes that derive RegexDecisionFilter and RegexDecisionFilterTest from SkipDocumentFilter and SkipDocumentFilterTest.

### DIFF
--- a/src/com/google/enterprise/adaptor/prebuilt/RegexDecisionFilter.java
+++ b/src/com/google/enterprise/adaptor/prebuilt/RegexDecisionFilter.java
@@ -72,7 +72,7 @@ import java.util.regex.Pattern;
    metadata.transform.pipeline.skipDocumentFilter.corpora=metadata
    </code></pre>
  */
-public class SkipDocumentFilter implements MetadataTransform {
+public class RegexDecisionFilter implements MetadataTransform {
   /**
    * Which collections of keys/values to search.  Metadata, params, or both.
    */
@@ -103,7 +103,7 @@ public class SkipDocumentFilter implements MetadataTransform {
   };
 
   private static final Logger log
-      = Logger.getLogger(SkipDocumentFilter.class.getName());
+      = Logger.getLogger(RegexDecisionFilter.class.getName());
 
   /** The name of the key (either Metadata key or params key) to match. */
   private String propertyName;
@@ -121,13 +121,18 @@ public class SkipDocumentFilter implements MetadataTransform {
   private boolean skipOnMatch = true;
 
   /**
+   * The {@code TransmissionDecision} to be made.
+   */
+  private TransmissionDecision decision;
+
+  /**
    * If {@code METADATA}, only search the metadata for the specified key;
    * if {@code PARAMS}, only search the params for the specified key;
    * if {@code METADATA_OR_PARAMS}, search both.
    */
   private Corpora corpora = Corpora.METADATA_OR_PARAMS;
 
-  private SkipDocumentFilter(String propertyName, Pattern pattern,
+  private RegexDecisionFilter(String propertyName, Pattern pattern,
       boolean skipOnMatch, Corpora corpora) {
     this.propertyName = propertyName;
     this.pattern = pattern;
@@ -201,7 +206,7 @@ public class SkipDocumentFilter implements MetadataTransform {
         log.info("Skipping document " + docId + ", because we found a match in "
             + corpora);
         params.put(MetadataTransform.KEY_TRANSMISSION_DECISION,
-            TransmissionDecision.DO_NOT_INDEX.toString());
+            decision.toString());
       } else {
         log.fine("Not skipping document " + docId + ", because we did not find "
             + "a match in " + corpora);
@@ -214,7 +219,7 @@ public class SkipDocumentFilter implements MetadataTransform {
         log.info("Skipping document " + docId + ", because we did not find a "
             + "match in " + corpora);
         params.put(MetadataTransform.KEY_TRANSMISSION_DECISION,
-            TransmissionDecision.DO_NOT_INDEX.toString());
+            decision.toString());
       }
     }
   }
@@ -222,22 +227,25 @@ public class SkipDocumentFilter implements MetadataTransform {
   @Override
   public String toString() {
     StringBuilder sb = new StringBuilder();
-    sb.append("SkipDocumentFilter(");
+    sb.append("RegexDecisionFilter(");
     sb.append(propertyName);
     sb.append(", ");
     sb.append(pattern == null ? "[null]" : pattern.toString());
     sb.append(", ");
     sb.append(skipOnMatch);
     sb.append(", ");
+    sb.append(decision);
+    sb.append(", ");
     sb.append(corpora);
     sb.append(")");
     return "" + sb;
   }
 
-  public static SkipDocumentFilter create(Map<String, String> cfg) {
+  public static RegexDecisionFilter create(Map<String, String> cfg) {
     String propertyName;
     Pattern pattern = null;
     boolean skipOnMatch = true;
+    TransmissionDecision decision;
     Corpora corpora;
 
     propertyName = cfg.get("propertyName");
@@ -264,9 +272,12 @@ public class SkipDocumentFilter implements MetadataTransform {
     }
     log.config("skipOnMatch set to " + skipOnMatch);
 
+    decision = TransmissionDecision.from(cfg.get("decision"));
+    log.config("decision = " + decision);
+
     corpora = Corpora.from(cfg.get("corpora"));
     log.config("corpora set to " + corpora);
 
-    return new SkipDocumentFilter(propertyName, pattern, skipOnMatch, corpora);
+    return new RegexDecisionFilter(propertyName, pattern, skipOnMatch, corpora);
   }
 }
--- a/src/com/google/enterprise/adaptor/prebuilt/RegexDecisionFilter.java
+++ b/src/com/google/enterprise/adaptor/prebuilt/RegexDecisionFilter.java
@@ -249,8 +249,8 @@ public class RegexDecisionFilter implements MetadataTransform {
     TransmissionDecision decision;
     Corpora corpora;
 
-    key = cfg.get("key");
-    if (Strings.isNullOrEmpty(key)) {
+    key = getTrimmedValue(cfg, "key");
+    if (key == null) {
       throw new NullPointerException("key may not be null or empty");
     }
     log.config("key = " + key);
@@ -264,21 +264,24 @@ public class RegexDecisionFilter implements MetadataTransform {
       log.config("pattern set to " + patternString);
     }
 
-    String decideOnMatchString = cfg.get("decideOnMatch");
+    String decideOnMatchString = getTrimmedValue(cfg, "decideOnMatch");
     if (decideOnMatchString != null) {
-      decideOnMatchString = decideOnMatchString.trim();
-    }
-    if (!Strings.isNullOrEmpty(decideOnMatchString)) {
       decideOnMatch = Boolean.parseBoolean(decideOnMatchString);
     }
     log.config("decideOnMatch set to " + decideOnMatch);
 
-    decision = TransmissionDecision.from(cfg.get("decision"));
+    decision = TransmissionDecision.from(getTrimmedValue(cfg, "decision"));
     log.config("decision = " + decision);
 
-    corpora = Corpora.from(cfg.get("corpora"));
+    corpora = Corpora.from(getTrimmedValue(cfg, "corpora"));
     log.config("corpora set to " + corpora);
 
-    return new RegexDecisionFilter(key, pattern, decideOnMatch, decision, corpora);
+    return new RegexDecisionFilter(key, pattern, decideOnMatch, decision,
+        corpora);
+  }
+
+  private static String getTrimmedValue(Map<String, String> cfg, String key) {
+    String value = cfg.get(key);
+    return (value == null) ? value : Strings.emptyToNull(value.trim());
   }
 }
--- a/src/com/google/enterprise/adaptor/prebuilt/RegexDecisionFilter.java
+++ b/src/com/google/enterprise/adaptor/prebuilt/RegexDecisionFilter.java
@@ -227,19 +227,13 @@ public class RegexDecisionFilter implements MetadataTransform {
 
   @Override
   public String toString() {
-    StringBuilder sb = new StringBuilder();
-    sb.append("RegexDecisionFilter(");
-    sb.append(key);
-    sb.append(", ");
-    sb.append(pattern == null ? "[null]" : pattern.toString());
-    sb.append(", ");
-    sb.append(decideOnMatch);
-    sb.append(", ");
-    sb.append(decision);
-    sb.append(", ");
-    sb.append(corpora);
-    sb.append(")");
-    return "" + sb;
+    return new StringBuilder("RegexDecisionFilter(")
+        .append(key).append(", ")
+        .append(pattern == null ? "[null]" : pattern.toString()).append(", ")
+        .append(decideOnMatch).append(", ")
+        .append(decision).append(", ")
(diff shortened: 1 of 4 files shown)


src/com/google/enterprise/adaptor/prebuilt/RegexDecisionFilter.java
```diff
@@ -0,0 +1,285 @@
...
+   metadata.transform.pipeline=regexDecisionFilter
+   metadata.transform.pipeline.regexDecisionFilter.factoryMethod=com.google.enterprise.adaptor.prebuilt.RegexDecisionFilter.create
+   metadata.transform.pipeline.regexDecisionFilter.key=Classification
+   metadata.transform.pipeline.regexDecisionFilter.pattern=(PUBLIC)|(DECLASSIFIED)
+   metadata.transform.pipeline.regexDecisionFilter.decideOnMatch=false
+   metadata.transform.pipeline.regexDecisionFilter.decision=do-not-index
+   metadata.transform.pipeline.regexDecisionFilter.corpora=metadata
+   </code></pre>
+ */
+public class RegexDecisionFilter implements MetadataTransform {
```
wiarlawd: I think decision is redundant, and it's not used elsewhere. The naming scheme we had was that a filter is a transform that makes a TransmissionDecision.

src/com/google/enterprise/adaptor/prebuilt/RegexDecisionFilter.java
```diff
@@ -0,0 +1,285 @@
...
+    public String toString() {
+      return name;
+    }
+  };
+
+  private static final Logger log
+      = Logger.getLogger(RegexDecisionFilter.class.getName());
+
+  /** The name of the key (either Metadata key or params key) to match. */
+  private String key;
```
wiarlawd: This rename from propertyName is for consistency with the PrebuiltTransforms? If we're doing that, can we rename corpora to something better?

src/com/google/enterprise/adaptor/prebuilt/RegexDecisionFilter.java
```diff
@@ -0,0 +1,285 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.enterprise.adaptor.prebuilt;
+
+import static com.google.enterprise.adaptor.MetadataTransform.TransmissionDecision;
+
+import com.google.common.base.Strings;
```
wiarlawd: PJ's preference is to avoid using Guava.

src/com/google/enterprise/adaptor/prebuilt/RegexDecisionFilter.java
```diff
@@ -255,18 +262,24 @@ public static SkipDocumentFilter create(Map<String, String> cfg) {
       log.config("pattern set to " + patternString);
     }
 
-    String skipOnMatchString = cfg.get("skipOnMatch");
-    if (skipOnMatchString != null) {
-      skipOnMatchString = skipOnMatchString.trim();
-    }
-    if (!Strings.isNullOrEmpty(skipOnMatchString)) {
-      skipOnMatch = Boolean.parseBoolean(skipOnMatchString);
+    String decideOnMatchString = getTrimmedValue(cfg, "decideOnMatch");
```
wiarlawd: PJ wants a better name here. What about "when" or "result", with values of "found" and "not-found" (which is actually what the code does; it uses find() rather than matches()):

pattern = 666
when = found
decision = do-not-index

src/com/google/enterprise/adaptor/prebuilt/RegexDecisionFilter.java
```diff
@@ -0,0 +1,285 @@
...
+   </code></pre>
+ */
+public class RegexDecisionFilter implements MetadataTransform {
+  /**
+   * Which collections of keys/values to search.  Metadata, params, or both.
+   */
+  private enum Corpora {
+    METADATA("metadata"),
+    PARAMS("params"),
+    METADATA_OR_PARAMS("metadata or params");
```
wiarlawd: This enum's name doesn't match all the others (spaces instead of dashes). I think it was basically just intended to be used as a default, and never explicitly. Maybe we could either rename this to ANYWHERE("anywhere") or add multiple names for the lookup: METADATA_OR_PARAMS("metadata or params", "anywhere")?

Similarly, maybe we could add "index" as a valid synonym for "as-is", either here, or in TransmissionDecision itself.

BrettMichaelJohnson: My goal is to move this enum into Metadata.java, rather than have it replicated in 4 separate transforms.  When I do that, we can consider making it smarter.

src/com/google/enterprise/adaptor/prebuilt/RegexDecisionFilter.java
BrettMichaelJohnson: Agreed

src/com/google/enterprise/adaptor/prebuilt/RegexDecisionFilter.java
wiarlawd: What about "corpora" => "location"? Also, "metadata or params" feels like a misfeature, but we can't maintain compatibility for SkipDocumentFilter without it. I think just "metadata" or "params" with a default of "metadata" would have been a better design. The default of either means you don't need to specify it as often, but you also open the door to random, unexpected behaviors.

src/com/google/enterprise/adaptor/prebuilt/RegexDecisionFilter.java
BrettMichaelJohnson: Done.

src/com/google/enterprise/adaptor/prebuilt/RegexFilter.java
```diff
@@ -69,13 +69,40 @@
...
+    private final String name;
+
+    private When(String name) {
+      this.name = name;
+    }
+
+    public static When from(String val) {
+      if ("not-found".equalsIgnoreCase(val)) {
+        return When.NOT_FOUND;
+      } else {
```
wiarlawd: This is a simple case where it is unlikely to arise (though we could add MATCHES), but being permissive like this means changing the behavior of customer configs if you want to add new values later.

src/com/google/enterprise/adaptor/prebuilt/RegexFilter.java
```diff
@@ -102,28 +103,19 @@ public String toString() {
...
-    public static Corpora from(String val) {
-      if ("metadata".equalsIgnoreCase(val)) {
-        return Corpora.METADATA;
-      }
-      if ("params".equalsIgnoreCase(val)) {
-        return Corpora.PARAMS;
-      }
-      return METADATA_OR_PARAMS;
+    public static Keyset from(String val) {
+      return (val == null) ? METADATA : Keyset.valueOf(val.toUpperCase());
```
wiarlawd: Add a US locale.

src/com/google/enterprise/adaptor/prebuilt/RegexFilter.java
```diff
@@ -102,28 +103,19 @@ public String toString() {
     }
   };
 
-  /**
-   * Which collections of keys/values to search.  Metadata, params, or both.
-   */
-  private enum Corpora {
+  // TODO (bmj): remove this and use the one in MetadataTransforms.
```
wiarlawd: Is that in this CL or planned for later? Need javadoc at some point, including a throws IllegalArgumentException tag for the from method.

src/com/google/enterprise/adaptor/prebuilt/RegexFilter.java
```diff
@@ -208,21 +199,20 @@ private boolean foundInParams(Map<String, String> params) {
...
       case METADATA:
         found = foundInMetadata(metadata);
         break;
       case PARAMS:
         found = foundInParams(params);
         break;
-      case METADATA_OR_PARAMS:
       default:
-        found = foundInMetadata(metadata) || foundInParams(params);
+        found = false;	// can't happen
```
wiarlawd: I feel like this should at least log a warning or even just throw AssertionError (the Google recommendation for enums), rather than just silently not match.

src/com/google/enterprise/adaptor/prebuilt/RegexFilter.java
```diff
@@ -261,26 +251,29 @@ public void transform(Metadata metadata, Map<String, String> params) {
   public String toString() {
     return new StringBuilder("RegexFilter(")
         .append(key).append(", ")
-        .append(pattern == null ? "[null]" : pattern.toString()).append(", ")
+        .append(keyset).append(", ")
+        .append(pattern.toString()).append(", ")
```
wiarlawd: toString is redundant.

src/com/google/enterprise/adaptor/prebuilt/RegexFilter.java
```diff
@@ -0,0 +1,299 @@
...
+    NOT_FOUND("not-found");
+
+    private final String name;
+
+    private When(String name) {
+      this.name = name;
+    }
+
+    public static When from(String val) {
+      if ("not-found".equalsIgnoreCase(val)) {
```
wiarlawd: You could think about using the same technique as Keyset, with replace('-', '_') added.

src/com/google/enterprise/adaptor/prebuilt/RegexFilter.java
```diff
@@ -0,0 +1,299 @@
...
+    public String toString() {
+      return name;
+    }
+  };
+
+  private static final Logger log
+      = Logger.getLogger(RegexFilter.class.getName());
+
+  /** The name of the key (either Metadata key or params key) to match. */
+  private String key;
```
wiarlawd: Can these fields all be final?

src/com/google/enterprise/adaptor/prebuilt/RegexFilter.java
```diff
@@ -0,0 +1,299 @@
...
+
+  /**
+   * If {@code METADATA}, search the metadata for the specified key;
+   * if {@code PARAMS}, search the params for the specified key;
+   */
+  private Keyset keyset = Keyset.METADATA;
```
wiarlawd: This value is not needed, since the factory method will always set this, and it prevents the use of final.

src/com/google/enterprise/adaptor/prebuilt/RegexFilter.java
```diff
@@ -0,0 +1,299 @@
...
+   * The regex pattern to match in the property value (can be null to indicate
+   * that any value is considered a match).
+   */
+  private Pattern pattern;
+
+  /**
+   * If {@code found}, make a transmission decision on a match;
+   * if {@code not-found}, make a transmission decision on a failed match.
+   */
+  private When when = When.FOUND;
```
wiarlawd: Same here.

src/com/google/enterprise/adaptor/prebuilt/RegexFilter.java
```diff
@@ -0,0 +1,299 @@
...
+        .append(pattern.toString()).append(", ")
+        .append(when).append(", ")
+        .append(decision).append(")")
+        .toString();
+  }
+
+  public static RegexFilter create(Map<String, String> cfg) {
+    String key;
+    Keyset keyset;
+    Pattern pattern = null;
```
wiarlawd: Shouldn't need null here, or maybe kill the C-style declarations?

wiarlawd (changes requested):

src/com/google/enterprise/adaptor/prebuilt/RegexFilter.java
```diff
@@ -0,0 +1,299 @@
...
+    boolean found;
+    switch (keyset) {
+      case METADATA:
+        found = foundInMetadata(metadata);
+        break;
+      case PARAMS:
+        found = foundInParams(params);
+        break;
+      default:
+        found = false;	// can't happen
```
wiarlawd: Remove tab character. Please check CLs for lint errors generally, there are half-dozen scattered around.

src/com/google/enterprise/adaptor/prebuilt/RegexFilter.java

src/com/google/enterprise/adaptor/prebuilt/RegexFilter.java

src/com/google/enterprise/adaptor/prebuilt/RegexFilter.java

src/com/google/enterprise/adaptor/prebuilt/RegexFilter.java

src/com/google/enterprise/adaptor/prebuilt/RegexFilter.java

src/com/google/enterprise/adaptor/prebuilt/RegexFilter.java
```diff
@@ -0,0 +1,320 @@
...
+
+    /** Returns the name of this {@code When} value. */
+    @Override
+    public String toString() {
+      return name;
+    }
+  };
+
+  // TODO (bmj): remove this and use the one in MetadataTransforms.
+  public enum Keyset {
```
PapaJoltOmega: s/public/private/

src/com/google/enterprise/adaptor/prebuilt/RegexFilter.java
```diff
@@ -0,0 +1,320 @@
...
+    METADATA("metadata"),
+    PARAMS("params");
```
PapaJoltOmega: does it make sense to have either/both as option?

src/com/google/enterprise/adaptor/prebuilt/RegexFilter.java
```diff
@@ -0,0 +1,320 @@
...
+    boolean found;
+    switch (keyset) {
+      case METADATA:
+        found = foundInMetadata(metadata);
+        break;
+      case PARAMS:
+        found = foundInParams(params);
+        break;
+      default:
+        found = false; // can't happen
```
PapaJoltOmega: instead of comment throw assertionerror

PapaJoltOmega (approved):
THank you

src/com/google/enterprise/adaptor/prebuilt/RegexFilter.java
```diff
@@ -0,0 +1,320 @@
...
+      }
+    }
+  }
+
+  @Override
+  public String toString() {
+    return new StringBuilder("RegexFilter(")
+        .append(key).append(", ")
+        .append(keyset).append(", ")
+        .append(pattern.toString()).append(", ")
```
wiarlawd: toString() is still redundant. ;-)

wiarlawd (commented):
This is not yet mergeable due to the code slush.

Bump a missed comment.

src/com/google/enterprise/adaptor/prebuilt/RegexFilter.java
```diff
@@ -227,7 +228,7 @@ public void transform(Metadata metadata, Map<String, String> params) {
         found = foundInParams(params);
         break;
       default:
-        found = false; // can't happen
+        throw new AssertionError("Invalid keyset");
```
wiarlawd: "Invalid keyset " + keyset

wiarlawd (approved):
LGTM, with a nit.

We're lifting the code slush and allowing commits on master.

PR #31, opened 2017-04-15
Allow connectors to add params for use by MetadataTransforms

BrettMichaelJohnson:
This change allows the connector to insert metadata associated
with the document into the params that MetadataTransforms
receive. This data may be useful to the transforms, but
will not be sent to the GSA.


PapaJoltOmega:
There are four commits in this pull request at this time. I think only the first commit was intended.  I'm thinking that currently this pull request is reflecting an incorrect range of git commits.  Please take a look.

BrettMichaelJohnson:
You are correct.  This seems to be based off of the groups branch instead of master.  I have rebased my local branch, but I cannot figure out how to get the pull request to do the right thing.  I will delete this one and create a new one.

PR #33, opened 2017-04-18
Allow connectors to add params for use by MetadataTransforms

BrettMichaelJohnson:
This change allows the connector to insert metadata associated
with the document into the params that MetadataTransforms
receive. This data may be useful to the transforms, but
will not be sent to the GSA.

### DIFF
--- a/src/com/google/enterprise/adaptor/CommandStreamParser.java
+++ b/src/com/google/enterprise/adaptor/CommandStreamParser.java
@@ -135,6 +135,13 @@ import java.util.regex.Pattern;
  * "meta-value=" -- specifies a metadata value associated with
  * immediately preceding metadata-name<p>
  *
+ * "param-name=" -- specifies a parameter key, to be followed by a parameter-value.
+ * Parameters are supplied to {@link MetadataTransforms} for use when making
+ * transforms or decisions.<p>
+ *
+ * "param-value=" -- specifies a parameter value associated with
+ * immediately preceding parameter-name<p>
+ *
  * "content" -- signals the beginning of binary content which
  * continues to the end of the file or stream<p>
  *
@@ -299,6 +306,8 @@ public class CommandStreamParser {
     MIME_TYPE("mime-type"),
     META_NAME("meta-name"),
     META_VALUE("meta-value"),
+    PARAM_NAME("param-name"),
+    PARAM_VALUE("param-value"),
     CONTENT("content"),
     AUTHZ_STATUS("authz-status"),
     SECURE("secure"),
@@ -482,6 +491,17 @@ public class CommandStreamParser {
               new Object[] {docId.getUniqueId(), metaName,
                 command.getArgument()});
           response.addMetadata(metaName, command.getArgument());
+          break;
+        case PARAM_NAME:
+          String paramName = command.getArgument();
+          command = readCommand();
+          if (command == null || command.getOperation() != Operation.PARAM_VALUE) {
+            throw new IOException("param-name must be immediately followed by param-value");
+          }
+          log.log(Level.FINEST, "Retriever: {0} has parameter {1}={2}",
+              new Object[] {docId.getUniqueId(), paramName,
+                command.getArgument()});
+          response.addParam(paramName, command.getArgument());
           break;
         case UP_TO_DATE:
           log.log(Level.FINEST, "Retriever: {0} is up to date.", docId.getUniqueId());
--- a/src/com/google/enterprise/adaptor/CommandStreamParser.java
+++ b/src/com/google/enterprise/adaptor/CommandStreamParser.java
@@ -501,7 +501,7 @@ public class CommandStreamParser {
           log.log(Level.FINEST, "Retriever: {0} has parameter {1}={2}",
               new Object[] {docId.getUniqueId(), paramName,
                 command.getArgument()});
-          response.addParam(paramName, command.getArgument());
+          response.setParam(paramName, command.getArgument());
           break;
         case UP_TO_DATE:
           log.log(Level.FINEST, "Retriever: {0} is up to date.", docId.getUniqueId());
--- a/src/com/google/enterprise/adaptor/DocumentHandler.java
+++ b/src/com/google/enterprise/adaptor/DocumentHandler.java
@@ -45,7 +45,6 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.Date;
-import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Locale;
@@ -734,6 +733,7 @@ class DocumentHandler implements HttpHandler {
     private boolean crawlOnce;
     private boolean lock;
     private Map<String, Acl> fragments = new TreeMap<String, Acl>();
+    private Map<String, String> params = new TreeMap<String, String>();
 
     public DocumentResponse(HttpExchange ex, DocId docId, Thread thread) {
       this.ex = ex;
@@ -991,6 +991,15 @@ class DocumentHandler implements HttpHandler {
         throw new IllegalStateException("Already responded");
       }
       this.lock = lock;
+
+    }
+
+    @Override
+    public void addParam(String key, String value) {
+      if (state != State.SETUP) {
+        throw new IllegalStateException("Already responded");
+      }
+      params.put(key, value);
     }
 
     private long getWrittenContentSize() {
@@ -1213,7 +1222,6 @@ class DocumentHandler implements HttpHandler {
         log.log(Level.FINER, "Not performing Metadata transform.");
         return;
       }
-      Map<String, String> params = new HashMap<String, String>();
       params.put(KEY_DOC_ID, docId.getUniqueId());
       params.put(KEY_CONTENT_TYPE, finalContentType);
       if (null != lastModified) {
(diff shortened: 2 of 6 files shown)


wiarlawd (commented):
I don't see the utility of this. You don't eliminate the risk of name collisions, you just risk them with the library instead of the repository. Is there something you could do with this that you can't do by creating and deleting a metadata element for the same purpose?

John L

src/com/google/enterprise/adaptor/Response.java
```diff
@@ -213,4 +213,18 @@
...
+   * but might not be indexed and searchable. The {@code params} include the
+   * documents {@link DocId}, and values from {@link setLock},
+   * {@link setCrawlOnce}, {@code setDisplayUrl}, {@code setContentType},
+   * and {@code setLastModified}.
+   *
+   * @param key a key for a Map entry
+   * @param value the value for the Map entry for key
+   */
+  // TODO (bmj): Supply Params to ContentTransforms.
+  public void addParam(String key, String value);
```
wiarlawd: Just for the record, this has the same problem that the added pushGroupDefinitions overload has, breaking source compatibility with unit tests.

As the basis for stickiness, I would much prefer a setTransmissionDecision or setDoNotSkip method, rather than opening the door to adding random params that the library may or may not care about. I think the library should validate the params that are passable here, if you want to retain the generic method. You also can't overwrite existing library params with the current code, but that isn't documented here.

src/com/google/enterprise/adaptor/DocumentHandler.java
```diff
@@ -1243,7 +1251,7 @@ private void transformMetadata() {
       }
       try {
         final String du = params.get(KEY_DISPLAY_URL);
-        if (!Strings.isNullOrEmpty(du) && !du.equals(origDisplayUrlStr)) {
+        if (!Strings.isNullOrEmpty(du)) {
```
wiarlawd: Why is this changing here?

src/com/google/enterprise/adaptor/WrapperAdaptor.java
```diff
@@ -407,6 +418,10 @@ public boolean isCrawlOnce() {
     public boolean isLock() {
       return lock;
     }
+
+    public Map<String, String> getParams() {
+      return params;
```
wiarlawd, line 423: Return a copy? I know this is just "like the others", but what's it for?

src/com/google/enterprise/adaptor/CommandStreamParser.java
```diff
@@ -483,6 +492,17 @@ public void readFromRetriever(DocId docId, Response response) throws IOException
                 command.getArgument()});
           response.addMetadata(metaName, command.getArgument());
           break;
+        case PARAM_NAME:
```
wiarlawd, line 495: Did you actually try running whatever this is?

wiarlawd (changes requested):
I missed this as the underpinning for stickiness until I read the other CL.

John L

BrettMichaelJohnson:
As an example of why this is useful, the Filesystem adaptor wants to do this:

    // Add the rest of the file attributes into the params for use by           
    // MetadataTransforms.                                                      
    resp.addParam("Creation-Date", dateFormatter.get().format(
        new Date(attrs.creationTime().toMillis())));
    resp.addParam("Last-Modified-Date", dateFormatter.get().format(
        new Date(attrs.lastModifiedTime().toMillis())));
    resp.addParam("Last-Access-Date", dateFormatter.get().format(
        new Date(attrs.lastAccessTime().toMillis())));
    resp.addParam("Is-Directory", Boolean.toString(attrs.isDirectory()));
    resp.addParam("Is-Regular-File", Boolean.toString(attrs.isRegularFile()));
    if (attrs.isRegularFile()) {
      rep.addParam("File-Size", Long.toString(attrs.size()));
    }

This make the BasicFileAttributes available to the transforms for decision
and/or copying into metadata to be indexed.  For instance the dates may
be used by the DateFilter (deprecating the internal one).  And original 
file size is b/18996895.

src/com/google/enterprise/adaptor/DocumentHandler.java
BrettMichaelJohnson: I honestly cannot remember why I thought this was needed.

src/com/google/enterprise/adaptor/CommandStreamParser.java
BrettMichaelJohnson: I ran its test.

src/com/google/enterprise/adaptor/WrapperAdaptor.java
BrettMichaelJohnson: This class should be marked VisibleForTesting.  It is used by CommandStreamParserTest and others that use TestHelper.  Basically the getters allow the tests to extract fields from the Response.

wiarlawd:
I'm just barely buying the argument about metadata that might be useful to transforms, but which the connector does not want to decide to index on its own. Access date is the classic example.

What about allowing extended param names only with "X-" prefixes, to avoid collisions?

It's a little weird that addParam (maybe putParam, since it's a map and multiple values are not supported?) is a duplicate of setLock, setDisplayUrl, etc., but that Forced-Transmission-Decision requires you to use addParam.

BrettMichaelJohnson:
I was thinking of strongly suggested namespaces for the params, i.e. FS_LAST_ACCESS_DATE.
I can accept the dichotomy between params that get sent to the GSA (lock, displayUrl, etc) and params that are used by the filters.  Adding a transmissionDecision/sticky setter to the response wouldn't be too terrible, but I really see it as six-of-a-half-dozen.

wiarlawd:
I don't see a need for a per-connector type namespace, the scope here is just one connector instance. But I would prefer an enforced namespace.

If we keep addParam for Forced-Transmission-Decision, then I think we should choose between adding setForcedTransmissionDecision, too, which weakens addParam as a feature, or deprecating setDisplayUrl, setLock, et al. (if there are others).

src/com/google/enterprise/adaptor/DocumentHandler.java
wiarlawd: It's not your code, but it just prevents us from reparsing the URL if it was changed by the transform pipeline. Seems harmless enough and not worth changing.

src/com/google/enterprise/adaptor/DocumentHandler.java
```diff
@@ -991,6 +991,15 @@ public void setLock(boolean lock) {
         throw new IllegalStateException("Already responded");
       }
       this.lock = lock;
+
```
wiarlawd: Delete blank line.

src/com/google/enterprise/adaptor/DocumentHandler.java
```diff
@@ -991,6 +991,15 @@ public void setLock(boolean lock) {
...
+    }
+
+    @Override
+    public void addParam(String key, String value) {
```
wiarlawd: I would go with putParam (collection semantics) or setParam (generic meaning), rather than addParam, which suggests multiple values are possible.

wiarlawd:
Just as a reminder, to myself as much as anyone, Brett and I had an offline chat about namespaces, and agreed to require an X- prefix for names not matching library keys. In fact, I would go as far as adding setForcedTransmissionDecision to the other CL (the alternative being to deprecate the other existing setters), and having the generic addParam method flat require the X- prefix, that is, you could not set the predefined library keys with this method. That also means that this CL is not a prerequisite to #37, but an independent feature, justified mainly by access time (and some other odd metadata where indexing doesn't make much sense but filtering does).

wiarlawd:
Please add a test that at least calls addParam, maybe something like adding access time with a filter that skips the document?

wiarlawd (changes requested):

wiarlawd:
This change has javadoc errors in {@link MetadataTransforms} in two files.

src/com/google/enterprise/adaptor/DocumentHandler.java
BrettMichaelJohnson: I choose setParam()

src/com/google/enterprise/adaptor/DocumentHandler.java

src/com/google/enterprise/adaptor/DocumentHandler.java
BrettMichaelJohnson: Restored.

src/com/google/enterprise/adaptor/Response.java
BrettMichaelJohnson: I enforced the param key must start with "X-"

test/com/google/enterprise/adaptor/DocumentHandlerTest.java
```diff
@@ -767,6 +767,41 @@ public void getDocContent(Request request, Response response)
...
+    List<MetadataTransform> transforms = new LinkedList<MetadataTransform>();
+    transforms.add(new MetadataTransform() {
+      @Override
+      public void transform(Metadata metadata, Map<String, String> params) {
+        if (params.containsKey("X-Do-Not-Index")) {
+          params.put("Transmission-Decision", "do-not-index");
+        }
+      }
+    });
+    MetadataTransformPipeline transform = new MetadataTransformPipeline(transforms,
```
wiarlawd: 80 cols

wiarlawd (approved):
LGTM, with a nit.

Please wait for PJ to review.

PapaJoltOmega (commented):
To sum up the review thread:
- This feature is for allowing someone to filter on acces time and other values that change more frequently than indexed metadata
- This feature is no longer being made to support sticky transform decision

Yes?

Thank you

BrettMichaelJohnson:
PJ,
Yes. We chose another method of setting the sticky transform decision.

PapaJoltOmega (approved):

PR #34, opened 2017-04-20
Add Millisecond Date support to DateFilter

BrettMichaelJohnson:
Last modified time is supplied as milliseconds since the epoch UTC
in the Params, so this adds a MillisecondDateFormat that can parse it.

### DIFF
--- a/src/com/google/enterprise/adaptor/prebuilt/DateFilter.java
+++ b/src/com/google/enterprise/adaptor/prebuilt/DateFilter.java
@@ -16,12 +16,16 @@ package com.google.enterprise.adaptor.prebuilt;
 
 import static com.google.enterprise.adaptor.MetadataTransform.TransmissionDecision;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
 import com.google.common.base.Strings;
 import com.google.enterprise.adaptor.Metadata;
 import com.google.enterprise.adaptor.MetadataTransform;
 
+import java.text.DateFormat;
+import java.text.FieldPosition;
 import java.text.ParseException;
+import java.text.ParsePosition;
 import java.text.SimpleDateFormat;
 import java.util.Date;
 import java.util.Map;
@@ -39,15 +43,19 @@ import java.util.logging.Logger;
  * {@code key} The name of the metadata or param property whose date value
  * determines whether the document will be skipped or not.
  * <p>
- * {@code format} The {@link SimpleDateFormat} used to parse the document's
+ * {@code format} The {@link DateFormat} used to parse the document's
  * date values. If no {@code format} is specified, a lenient ISO8601 format
- * ("yyyy-MM-dd") is used.
+ * ("yyyy-MM-dd") is used. If format is "millis", then the dates
+ * are parsed as if they were milliseconds since the epoch
+ * (January 1, 1970, 00:00:00 GMT). Otherwise, the format must parsable
+ * by {@link SimpleDateFormat}.
  * <p>
  * {@code date} The cut-off for the date value. Document's whose date
  * value is before the configued {@code date} will not be indexed.
  * The configured {@code date} must be parsable by the configured date
- * {@code format}. Only one of {@code date} or {@code days} configuration
- * may be specified.
+ * {@code format}, unless {@code format} is "millis", in which
+ * case the date must be specified in ISO8601 format. Only one of
+ * {@code date} or {@code days} configuration may be specified.
  * <p>
  * {@code days} The cut-off for the date value. Document's whose date
  * value is more than {@code days} before present will not be indexed.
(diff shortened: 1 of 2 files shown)


src/com/google/enterprise/adaptor/prebuilt/DateFilter.java
```diff
@@ -241,7 +254,8 @@ public static DateFilter create(Map<String, String> cfg) {
       format = ISO_8601_FORMAT;
     }
     log.config("format = " + format);
-    SimpleDateFormat dateFormat = new SimpleDateFormat(format);
+    SimpleDateFormat dateFormat = new SimpleDateFormat(
+        "millis".equalsIgnoreCase(format) ? ISO_8601_FORMAT : format);
```
PapaJoltOmega: If config's format is "millis" then the internal format is "yyyy-MM-dd"? Please add comment explaining.

src/com/google/enterprise/adaptor/prebuilt/DateFilter.java
```diff
@@ -241,7 +254,8 @@ public static DateFilter create(Map<String, String> cfg) {
       format = ISO_8601_FORMAT;
     }
     log.config("format = " + format);
-    SimpleDateFormat dateFormat = new SimpleDateFormat(format);
+    SimpleDateFormat dateFormat = new SimpleDateFormat(
```
PapaJoltOmega: Only use seems to be on 270. Can we put declaration there?

PapaJoltOmega (commented):
Thank you

src/com/google/enterprise/adaptor/prebuilt/DateFilter.java
BrettMichaelJohnson: Comment is on line 56.

src/com/google/enterprise/adaptor/prebuilt/DateFilter.java
BrettMichaelJohnson: This is intentional, so that invalid format strings throw exception in create(), whether you use "date" or "days".

src/com/google/enterprise/adaptor/prebuilt/DateFilter.java
```diff
@@ -39,15 +43,19 @@
  * {@code key} The name of the metadata or param property whose date value
  * determines whether the document will be skipped or not.
  * <p>
- * {@code format} The {@link SimpleDateFormat} used to parse the document's
+ * {@code format} The {@link DateFormat} used to parse the document's
  * date values. If no {@code format} is specified, a lenient ISO8601 format
- * ("yyyy-MM-dd") is used.
+ * ("yyyy-MM-dd") is used. If format is "millis", then the dates
```
PapaJoltOmega: even if format is specified lenient is used.  only if millis is specified then lenient is not used.

src/com/google/enterprise/adaptor/prebuilt/DateFilter.java
```diff
@@ -39,15 +43,19 @@
...
+ * by {@link SimpleDateFormat}.
  * <p>
  * {@code date} The cut-off for the date value. Document's whose date
  * value is before the configued {@code date} will not be indexed.
  * The configured {@code date} must be parsable by the configured date
- * {@code format}. Only one of {@code date} or {@code days} configuration
- * may be specified.
+ * {@code format}, unless {@code format} is "millis", in which
+ * case the date must be specified in ISO8601 format. Only one of
+ * {@code date} or {@code days} configuration may be specified.
```
PapaJoltOmega, line 58: Please insert example configs. One with millis that  would work with last_modified param.

src/com/google/enterprise/adaptor/prebuilt/DateFilter.java
PapaJoltOmega: How about adding "// Even when considered per doc values are millis config's cutoffs are ISO_8601"

PapaJoltOmega (approved):
thank you

src/com/google/enterprise/adaptor/prebuilt/DateFilter.java
BrettMichaelJohnson: Some wordsmithing done.

src/com/google/enterprise/adaptor/prebuilt/DateFilter.java
BrettMichaelJohnson: Done.

src/com/google/enterprise/adaptor/prebuilt/DateFilter.java
BrettMichaelJohnson: Done.

PapaJoltOmega (approved):
Great. Thank you. LGTM

PR #37, opened 2017-04-21
Add a forced transmission decision that should not be overridden

BrettMichaelJohnson:
Graph traversal connectors break when folders are skipped,
so they need a way to override filters that may attemt to
skip these folders. This change adds a Forced Transmission
Decision that overrides any Transmission Decision set by
the filters. As a convenience, our built-in filters defer
to a forced decision and return immediately.


PapaJoltOmega (commented):
If someone had a transform at end of transform pipeline that said AS_IS that would be equivalent to having sticky decision earlier in pipeline. The difference being the early termination of sticky decision?  Is that right?

BrettMichaelJohnson:
Yes, a final AS_IS would work.  The problem this is addressing is unintentional breakage of graph traversers through poorly configured or improperly configured transform pipeline. The current filters are not sophisticated enough to be discriminant in their dropping of documents.  Imagine skipping docs based on date.  The root of a filesystem likely hasn't been modified in years, so dropping the start path immediately halts traversal.

Putting things back is one way to address it.  The problem is determining which items are folders/containers for graph traversals from a MetadataTransform ranges from trivial for FS connector (docid ends in /), to difficult with Opentext (one of a dozen or so object types), to nearly impossible for Documentum (determining if a custom object derives from dm_folder).

The connectors themselves know which items need to be protected, so this feature + the addParam CL, provide a mechanism for the connectors to tag crucial items.  The "Forced" implementation also leaves the option of in the future adding something like a --force configuration to a filter that can override other decisions (say you wanted to intentionally prune a tree, even though you can configure that on the GSA with a do-not-crawl pattern).

wiarlawd (commented):
You basically already implemented --force, and then modified each prebuilt filter to prevent it's use, though custom transforms could still do so. Allowing --force is the only reason I can think of for the sticky flag to be in the params in the first place.

John L

src/com/google/enterprise/adaptor/prebuilt/DateFilter.java
```diff
@@ -221,8 +221,20 @@ private String excludedByDateInParams(Map<String, String> params) {
    */
   @Override
   public void transform(Metadata metadata, Map<String, String> params) {
-    String excludedDate;
+    String docId = params.get(MetadataTransform.KEY_DOC_ID);
+    if (null == docId || docId.isEmpty()) {
+      docId = "with no docId";
+    }
+
+    String forced
```
wiarlawd: I still don't like this, though it's not as bad as I originally thought. It can't see the entire pipeline, so it doesn't know what's going to happen downstream. The force flag is not the only mechanism to override a forced transmission decision. The copy transform can do it, too (or will be able to, with #42).

src/com/google/enterprise/adaptor/DocumentHandler.java
```diff
@@ -1253,7 +1254,12 @@ private void transformMetadata() {
       crawlOnce = Boolean.parseBoolean(params.get(KEY_CRAWL_ONCE));
       lock = Boolean.parseBoolean(params.get(KEY_LOCK));
       // TODO: make constants for this growing set of valid keys
-      considerNotSending(params.get(KEY_TRANSMISSION_DECISION), docId);
+      String transmissionDecision
+          = params.get(KEY_FORCED_TRANSMISSION_DECISION);
+      if (Strings.isNullOrEmpty(transmissionDecision)) {
```
wiarlawd: PJ has asked us to stop using Guava in the adaptor library.

wiarlawd: The empty string would lead to an IllegalArgumentException in KEY_TRANSMISSION_DECISION, and I think it should behave the same for KEY_FORCED_TRANSMISSION_DECISION. That is, this really should be just transmissionDecision == null anyway.

test/com/google/enterprise/adaptor/DocumentHandlerTest.java
```diff
@@ -698,6 +698,42 @@ public void getDocContent(Request request, Response response)
   }
 
   @Test
+  public void testForcedDecision() throws Exception {
+    List<MetadataTransform> transforms = new LinkedList<MetadataTransform>();
+    transforms.add(new MetadataTransform() {
+      @Override
+      public void transform(Metadata metadata, Map<String, String> params) {
+        params.put("Forced-Transmission-Decision", "as-is");
```
wiarlawd: Why did you do this here rather than all addParam (or setForcedTransmissionDecision or whatever) in getDocContent?

wiarlawd:
Recording an offline chat for posterity. Brett and I agreed to implement force now, rather than waiting, although as I mentioned in #33, you can configure a copy transform to implement the force flag (i.e., override the Forced-Transmission-Decision param).

I would like to add a method here for connectors to use, rather than the generic addParam (which needlessly conflicts with a handful of existing methods to set other params).

test/com/google/enterprise/adaptor/DocumentHandlerTest.java
```diff
@@ -698,6 +698,42 @@ public void getDocContent(Request request, Response response)
...
+        params.put("Transmission-Decision", "do-not-index");
+        metadata.set("docid", params.get("DocId"));
+      }
+    });
+    MetadataTransformPipeline transform = new MetadataTransformPipeline(transforms,
```
wiarlawd: 80 cols

wiarlawd (changes requested):
Apparently forgot to send this some time back.

I think I mentioned this somewhere, but please add setForcedTransmissionDecision to Response.

src/com/google/enterprise/adaptor/DocumentHandler.java

src/com/google/enterprise/adaptor/prebuilt/DateFilter.java
BrettMichaelJohnson: Reverted.

src/com/google/enterprise/adaptor/prebuilt/FilterMimetypes.java
```diff
@@ -119,6 +120,19 @@ private void split(Set<String> globs, Set<String> explicit, Set<String> src) {
...
+    }
+
+    String forced
+        = params.get(MetadataTransform.KEY_FORCED_TRANSMISSION_DECISION);
+    if (!Strings.isNullOrEmpty(forced)) {
+      log.log(Level.FINE, "Not evaluating document {0}  because transmission "
+          + "decision {1} is forced.", new Object[] { docId, forced });
+      return;
+    }
+
```
BrettMichaelJohnson: Reverted.

src/com/google/enterprise/adaptor/prebuilt/SkipDocumentFilter.java
```diff
@@ -178,6 +178,19 @@ private boolean foundInParams(Map<String, String> params) {
...
+    }
+
+    String forced
+        = params.get(MetadataTransform.KEY_FORCED_TRANSMISSION_DECISION);
+    if (!Strings.isNullOrEmpty(forced)) {
+      log.fine("Not evaluating document " + docId + " because transmission "
+          + "decision " + forced + " is forced.");
+      return;
+    }
+
```
BrettMichaelJohnson: Reverted

test/com/google/enterprise/adaptor/DocumentHandlerTest.java
BrettMichaelJohnson: Done.

test/com/google/enterprise/adaptor/DocumentHandlerTest.java
BrettMichaelJohnson: Neither was in place at the time, but now I setForcedTransmissionDecision in the response.

src/com/google/enterprise/adaptor/DocumentHandler.java
```diff
@@ -1253,7 +1268,12 @@ private void transformMetadata() {
       crawlOnce = Boolean.parseBoolean(params.get(KEY_CRAWL_ONCE));
       lock = Boolean.parseBoolean(params.get(KEY_LOCK));
       // TODO: make constants for this growing set of valid keys
```
wiarlawd: Remove this obsolete TODO, fixed in commit 84f79d9b.

src/com/google/enterprise/adaptor/Response.java
```diff
@@ -14,6 +14,8 @@
 
 package com.google.enterprise.adaptor;
 
+import static com.google.enterprise.adaptor.MetadataTransform.TransmissionDecision;
```
wiarlawd: Not a static import.

src/com/google/enterprise/adaptor/WrapperAdaptor.java
```diff
@@ -14,6 +14,8 @@
 
 package com.google.enterprise.adaptor;
 
+import static com.google.enterprise.adaptor.MetadataTransform.TransmissionDecision;
```
wiarlawd: Not a static import, though it will get deleted shortly anyway.

test/com/google/enterprise/adaptor/DocumentHandlerTest.java
```diff
@@ -14,6 +14,7 @@
 
 package com.google.enterprise.adaptor;
 
+import static com.google.enterprise.adaptor.MetadataTransform.TransmissionDecision;
```
wiarlawd: Yeah

test/com/google/enterprise/adaptor/prebuilt/CommandLineAdaptorTest.java
```diff
@@ -14,6 +14,7 @@
 
 package com.google.enterprise.adaptor.prebuilt;
 
+import static com.google.enterprise.adaptor.MetadataTransform.TransmissionDecision;
```
wiarlawd: Just yeah

wiarlawd (changes requested):

PapaJoltOmega (commented):
Is this really cleaner than having the connector do something like:

response.addMetadata("is_folder", "yes")

and then transform that says AS_IS for metadata key is_folder?

Let's please chat at our next meeting.  Thank you

src/com/google/enterprise/adaptor/DocumentHandler.java

src/com/google/enterprise/adaptor/Response.java
BrettMichaelJohnson: Done.

src/com/google/enterprise/adaptor/WrapperAdaptor.java
BrettMichaelJohnson: Done.

test/com/google/enterprise/adaptor/DocumentHandlerTest.java
BrettMichaelJohnson: Done.

test/com/google/enterprise/adaptor/prebuilt/CommandLineAdaptorTest.java
BrettMichaelJohnson: Done.

BrettMichaelJohnson:
PJ, your suggestion requires the administrator to configure another transform for the system to perform properly. One of the primary design goals of this change is to avoid accidental breakage through poor configuration.

wiarlawd (approved):

src/com/google/enterprise/adaptor/DocumentHandler.java
```diff
@@ -1252,8 +1267,12 @@ private void transformMetadata() {
       }
       crawlOnce = Boolean.parseBoolean(params.get(KEY_CRAWL_ONCE));
       lock = Boolean.parseBoolean(params.get(KEY_LOCK));
-      // TODO: make constants for this growing set of valid keys
-      considerNotSending(params.get(KEY_TRANSMISSION_DECISION), docId);
+      String transmissionDecision
+          = params.get(KEY_FORCED_TRANSMISSION_DECISION);
```
PapaJoltOmega, line 1282: So filters can override an already forced decision.  My only remaining thoughts are whether forced is the right adjective.  For example would coerced be preferable because it seems less final?

test/com/google/enterprise/adaptor/DocumentHandlerTest.java
```diff
@@ -698,6 +699,42 @@ public void getDocContent(Request request, Response response)
...
+        .setMetadataTransform(transform)
+        .build();
+    mockAdaptor.documentBytes = new byte[] {1, 2, 3};
+    handler.handle(ex);
+    assertEquals(200, ex.getResponseCode());
+    assertEquals("docid=test%20docId,test%20key=TESTING%20VALUE",
+        ex.getResponseHeaders().getFirst("X-Gsa-External-Metadata"));
+    // content remained as-is
+    assertArrayEquals(new byte[] {1, 2, 3}, ex.getResponseBytes());
+  }
```
PapaJoltOmega, line 735: could we add test where transform overrides the forced decision? that is intentionally available behaviour, yes?

PapaJoltOmega (approved):
Thank you

src/com/google/enterprise/adaptor/DocumentHandler.java
BrettMichaelJohnson: We are leaving open the possibility of supporting the equivalent of --force for filters to set FORCED_TRANSMISSION_DECISION instead of TRANSMISSION_DECISION.  We will try to keep you from shooting yourself in the foot, but that might be exactly what some customer wants to do.

test/com/google/enterprise/adaptor/DocumentHandlerTest.java
BrettMichaelJohnson: Done.

wiarlawd (approved):
LGTM again.

I peeked at the conflicts, and they are easy but I wanted to call out the second pair. The first pair are just keep both, you added setParam in another CL. The second pair are updates to code I deleted, so you just want to delete everything in your side of the conflicts, leaving just the "class WrapperAdaptor {" in one case and nothing in the other.

PR #38, opened 2017-04-27
Set the dist file suffix from the auto-detected version.

wiarlawd:
adaptorlib.version is auto-detected using git, and adaptorlib.suffix
is set automatically from the version. Each property can be overridden
independently from the command line.

Previously, adaptorlib.suffix was set before adaptorlib.version was
set, so you always had to set adaptorlib.suffix on the command line.
Also, if you set adaptorlib.version on the command line,
adaptorlib.suffix would incorrectly include the leading "v" from the
version number, if the standard "v1.2.3" format was used.

Now, most uses of the "package" target, including releases, can be
done without command line arguments.

When used without git, the auto-detected version used in the JAR
manifest was "unknown", and the default suffix was "-yyyyMMdd".
Now the version will be "unknown-yyyyMMdd" and the suffix will be
"-unknown-yyyyMMdd".

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -215,6 +215,14 @@
     </exec>
     <!-- Set version if git describe failed. -->
     <property name="adaptorlib.version" value="unknown-${DSTAMP}"/>
+
+    <!-- Use the version without the "v" prefix for file names. -->
+    <loadresource property="adaptorlib.suffix">
+      <propertyresource name="adaptorlib.version"/>
+      <filterchain>
+          <replaceregex pattern="^(v(\d))?" replace="-\2"/>
+      </filterchain>
+    </loadresource>
   </target>
 
   <target name="dist" description="Generate distribution binaries"
@@ -238,7 +246,7 @@
     <copy file="THIRDPARTYLICENSE.txt" todir="${dist.staging.dir}"/>
 
     <!-- adaptor.jar -->
-    <jar destfile="${dist.staging.dir}/adaptor-${adaptorlib.version}.jar"
+    <jar destfile="${dist.staging.dir}/adaptor${adaptorlib.suffix}.jar"
       basedir="${build-src.dir}" excludes="${adaptor.pkg.dir}/examples/**">
       <manifest>
         <section name="com/google/enterprise/adaptor/">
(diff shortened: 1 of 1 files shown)


wiarlawd:
Brett pointed out that the this adds a "v" prefix to the file name suffix. Previously, the JAR file manifest used the v, but not the file suffix. I'll take a look at that.

PapaJoltOmega:
Will there be a later update to http://googlegsa.github.io/adaptor/docs/ReleaseProcess.html ? (which is different repo)

build.xml
```diff
@@ -221,7 +214,15 @@
       <arg value="--always"/>
     </exec>
     <!-- Set version if git describe failed. -->
-    <property name="adaptorlib.version" value="unknown"/>
+    <property name="adaptorlib.version" value="unknown-${DSTAMP}"/>
```
PapaJoltOmega, line 217: so before and after:
if explciit version is given then that is used
if git describe gives version then that is used

however now if no version is given then version is "unknown-$time" where before it was just "unknown".

that seems ok.

build.xml
```diff
@@ -221,7 +214,15 @@
...
+
+    <!-- Use the version without the "v" prefix for file names. -->
+    <loadresource property="adaptorlib.suffix">
+      <propertyresource name="adaptorlib.version"/>
+      <filterchain>
+          <replaceregex pattern="^(v(\d))?" replace="-\2"/>
```
PapaJoltOmega: want to double check:

everything is going to match pattern on left because everything has a beginning and the rest of pattern is optional.

if version starts with "v321" then suffix will become "-321".

if version="unknown=2017" then suffix becomes .....  "-"? I'm not sure if I read that right, I'm guessing since 2nd grouping didn't match anything then we just get the "-".  Is that right reading?

PapaJoltOmega (commented):
Thank you. Couple questions.

build.xml
wiarlawd: Yeah, so version was "unknown" and the suffix was "$time", and now both are "unknown-$time".

build.xml
wiarlawd: Yes, we always get a leading -, and strip a leading v but only if it's followed by a digit. "v321" becomes "-321", "unknown-2017" becomes "-unknown-2017", and "voodoo" becomes "-voodoo".

wiarlawd (commented):
Yes, this is part of a few changes to ReleaseProcess.html.

build.xml
wiarlawd: I should say, "unknown" is pretty rare, because you basically have to not be using the repository with git, but changing the version to include the timestamp and the suffix to include a warning that we don't have a real version both seemed like mild improvements to a corner case.

PapaJoltOmega (approved):

PR #41, opened 2017-05-06
Add DateFilter

BrettMichaelJohnson:


src/com/google/enterprise/adaptor/prebuilt/DateFilter.java
```diff
@@ -0,0 +1,384 @@
...
+  };
+
+  /** The name of the key (either Metadata key or params key) to match. */
+  private final String key;
+
+  /**
+   * If {@code METADATA}, search the metadata for the specified key;
+   * if {@code PARAMS},  search the params for the specified key;
+   */
+  private Keyset keyset = Keyset.METADATA;
```
wiarlawd: As in RegexFilter, the create method will ensure that this is set to METADATA by default, so we can elide the value and make this field final.

test/com/google/enterprise/adaptor/prebuilt/DateFilterTest.java
```diff
@@ -0,0 +1,465 @@
...
+    assertEquals("DateFilter(lastModified, params, yyyy-MM-dd, 365 days)",
+        transform.toString());
+  }
+
+  @Test
+  public void testCreate_KeysetBogus() {
+    Map<String, String> config = new HashMap<String, String>();
+    config.put("key", "lastModified");
+    config.put("days", "365");
+    config.put("keyset", "bogus");
```
wiarlawd: Above and in other CLs, you moved keyset to just after key, but not here or the rest of the examples in this file.

test/com/google/enterprise/adaptor/prebuilt/DateFilterTest.java
```diff
@@ -0,0 +1,465 @@
...
+  @Test
+  public void testTransform_DateInMetadataAndParams() {
+    Map<String, String> config = new HashMap<String, String>();
+    config.put("key", "lastModified");
+    config.put("days", "365");
+    DateFilter transform = DateFilter.create(config);
+    Map<String, String> params = new HashMap<String, String>();
+    params.put("lastModified", dateFormat.format(new Date()));
+    Metadata metadata = new Metadata();
+    metadata.add("lastModified", "1977-06-18");
```
wiarlawd, line 449: Maybe flipping the dates would be better, because if "metadata or params" were still in play, it would always skip this document, but now it would not be skipped if the old date were in params.

src/com/google/enterprise/adaptor/prebuilt/DateFilter.java
```diff
@@ -0,0 +1,384 @@
...
+ * determines whether the document will be skipped or not.
+ * <p>
+ * {@code format} The {@link DateFormat} used to parse the document's
+ * date values. If no {@code format} is specified, a ISO8601 format
+ * ("yyyy-MM-dd") is used. If format is "millis", then the dates
+ * are parsed as if they were milliseconds since the epoch
+ * (January 1, 1970, 00:00:00 GMT). Otherwise, a lenient
+ * {@link SimpleDateFormat} with the format pattern will be used.
+ * <p>
+ * {@code date} The cut-off for the date value. Document's whose date
```
wiarlawd: s/Document's/Documents/

src/com/google/enterprise/adaptor/prebuilt/DateFilter.java
```diff
@@ -0,0 +1,384 @@
...
+ * value is before the configued {@code date} will not be indexed.
+ * The configured {@code date} must be parsable by the configured date
+ * {@code format}, unless {@code format} is "millis", in which
+ * case the date must be specified in ISO8601 format. Only one of
+ * {@code date} or {@code days} configuration may be specified.
+ * <p>
+ * {@code days} The cut-off for the date value. Document's whose date
```
wiarlawd: s/Document's/Documents/

wiarlawd (changes requested):
I would have appreciated a first commit of the originally deleted code, but I checked out the bmj-datefilter branch and diffed them from the command line.

My passing thought on the original design while I was on vacation was to add a comparison operator: before or after (chosen to match the methods on Date objects). Right now this class only supports before, essentially. The most obvious use case for after is to put the older documents into a separate collection on the GSA. We could certainly add it later if anyone asked for it.

You can implement "after" with this filter as-is, but it's a little cumbersome. Maybe there's an easier method, but here's what I came up with:

1. DateFilter some-date
2. RegexFilter Transmission-Decision do-not-index not-found foo
3. RegexFilter Transmission-Decision do-not-index found ""
4. RegexFilter Transmission-Decision foo found do-not-index

That is, change all empty values to foo, change all do-not-index values to empty, then change all foo values to do-index-index, basically flipping the decision. You could, of course, write a custom transform that explicitly flips the decision values.

John L

src/com/google/enterprise/adaptor/prebuilt/DateFilter.java
BrettMichaelJohnson: Done.

src/com/google/enterprise/adaptor/prebuilt/DateFilter.java
BrettMichaelJohnson: Done.

src/com/google/enterprise/adaptor/prebuilt/DateFilter.java
BrettMichaelJohnson: Done.

test/com/google/enterprise/adaptor/prebuilt/DateFilterTest.java
BrettMichaelJohnson: Done.

test/com/google/enterprise/adaptor/prebuilt/DateFilterTest.java
BrettMichaelJohnson: Fixed up the below test.

src/com/google/enterprise/adaptor/prebuilt/DateFilter.java
```diff
@@ -0,0 +1,391 @@
...
+ * </code></pre>
+ */
+public class DateFilter implements MetadataTransform {
+  private static final Logger log
+      = Logger.getLogger(DateFilter.class.getName());
+
+  private static final String ISO_8601_FORMAT = "yyyy-MM-dd";
+
+  // TODO (bmj): remove this and use the one in MetadataTransforms.
+  public enum Keyset {
```
PapaJoltOmega: private

PapaJoltOmega (approved):
Thank you

src/com/google/enterprise/adaptor/prebuilt/DateFilter.java
```diff
@@ -129,7 +126,7 @@ public String toString() {
    * If {@code METADATA}, search the metadata for the specified key;
    * if {@code PARAMS},  search the params for the specified key;
```
wiarlawd: Delete extra space after comma.

wiarlawd (approved):
LGTM, with a nit.

Not mergeable due to code slush.

PR #42, opened 2017-05-06
Fix b/36685309 Allow prebuilt filters to access params

BrettMichaelJohnson:


src/com/google/enterprise/adaptor/prebuilt/PrebuiltTransforms.java
```diff
@@ -14,6 +14,8 @@
 
 package com.google.enterprise.adaptor.prebuilt;
 
+import com.google.common.base.Strings;
```
wiarlawd: Really, stop doing this.

src/com/google/enterprise/adaptor/prebuilt/PrebuiltTransforms.java
```diff
@@ -14,6 +14,8 @@
...
+import com.google.common.collect.ImmutableSet;
```
wiarlawd: Unused, as far as I can tell.

src/com/google/enterprise/adaptor/prebuilt/PrebuiltTransforms.java
```diff
@@ -41,82 +44,132 @@
...
+   * be specified in the document's {@code Metadata}, but some key/values of
+   * interest (e.g. ContentType, DocId) exist in the document's {@code params}.
    *
    * <p>Example configuration:
-   * <pre><code>overwrite=false
-   *3.from=colour
-   *3.to=color
-   *5.from=author
-   *5.to=contributors</code></pre>
+   * <pre><code>
```
wiarlawd, line 107: Why this change?

src/com/google/enterprise/adaptor/prebuilt/PrebuiltTransforms.java
```diff
@@ -41,82 +44,132 @@
...
+    boolean overwrite
+        = Boolean.parseBoolean(getTrimmedValue(config, "overwrite"));
     List<KeyPairing> copies = parseCopies(config);
     if (copies.isEmpty()) {
       log.warning("No entries listed to be copied");
     }
     return new CopyTransform(copies, overwrite, false);
   }
 
+  @Deprecated
```
wiarlawd: These deprecated methods all need a javadoc tag saying that copy should be used instead. I'm not convinced it's a necessary change, either. These are MetadataTransforms, after all.

src/com/google/enterprise/adaptor/prebuilt/PrebuiltTransforms.java
```diff
@@ -205,26 +287,61 @@ public String toString() {
...
       dest = to;
     }
 
+    @Override
+    public String toString() {
+      return "(from=" + src + ",to=" + dest + ")";
+    }
+  }
+
+  /** Contains a key in keyset. */
```
wiarlawd: Maybe s/Contains/Describes/? s/in/in a/?

src/com/google/enterprise/adaptor/prebuilt/PrebuiltTransforms.java
```diff
@@ -205,26 +287,61 @@ public String toString() {
...
+    Key(String key, Keyset keyset) {
+      if (key == null || keyset == null) {
+        throw new NullPointerException();
+      }
+      this.key = key;
+      this.keyset = keyset;
+    }
+
+    @Override
+    public boolean equals(Object o) {
```
wiarlawd, line 333: You usually want to override hashCode if you override equals. I know this is a private class (now).

src/com/google/enterprise/adaptor/prebuilt/PrebuiltTransforms.java
```diff
@@ -335,48 +467,81 @@ public static MetadataTransform replaceMetadata(
...
+    }
+
+    private void replaceInParams(String key, Map<String, String> params) {
+      String original = getTrimmedValue(params, key);
+      if (original == null) {
+        log.log(Level.FINE, "No param value for {0}. Skipping", key);
+        return;
       }
+      log.log(Level.FINE,
+          "Replacing param value of {0} that match {1} with {2}: {3}",
```
wiarlawd: s/match/matches/, but this whole log message seems deceptive (along with its metadata predecessor). Something like "Replacing {1} with {2} in param value of {0}: {3}", and similarly for the metadata log message.

wiarlawd (changes requested):
Partial review. I haven't gotten to the tests yet.

John L

src/com/google/enterprise/adaptor/prebuilt/PrebuiltTransforms.java
```diff
@@ -41,82 +44,132 @@
...
    * configuration entries (where {@code X} is an integer). The value for each
-   * is a metadata key. Copies are applied in the increasing order of the
-   * integers.
+   * is a metadata or param  key. Copies are applied in the increasing order of
+   * the integers. The config keys {@code "X.from.keyset"} and
+   * {@code "X.to.keyset"} may be set to {@code metadata} or to {@code params}
+   * to restrict the source and destination to only {@code Metadata} or
+   * {@code params}, respectively.  Most keys/values of interest will normally
+   * be specified in the document's {@code Metadata}, but some key/values of
+   * interest (e.g. ContentType, DocId) exist in the document's {@code params}.
```
wiarlawd: s/ContentType/Content-Type/

This was copied from SkipDocumentFilter, and might need to be changed in RegexFilter, too.

src/com/google/enterprise/adaptor/prebuilt/PrebuiltTransforms.java
```diff
@@ -14,6 +14,8 @@
 
 package com.google.enterprise.adaptor.prebuilt;
 
+import com.google.common.base.Strings;
+import com.google.common.collect.ImmutableSet;
 import com.google.enterprise.adaptor.Metadata;
 import com.google.enterprise.adaptor.MetadataTransform;
```
wiarlawd: Collection is also an unused import.

src/com/google/enterprise/adaptor/prebuilt/PrebuiltTransforms.java
BrettMichaelJohnson: Done.

src/com/google/enterprise/adaptor/prebuilt/PrebuiltTransforms.java
BrettMichaelJohnson: Done.

src/com/google/enterprise/adaptor/prebuilt/PrebuiltTransforms.java
BrettMichaelJohnson: Done.

src/com/google/enterprise/adaptor/prebuilt/PrebuiltTransforms.java
BrettMichaelJohnson: Done.

src/com/google/enterprise/adaptor/prebuilt/PrebuiltTransforms.java
BrettMichaelJohnson: Done.

src/com/google/enterprise/adaptor/prebuilt/PrebuiltTransforms.java
BrettMichaelJohnson: Done.

src/com/google/enterprise/adaptor/prebuilt/PrebuiltTransforms.java
BrettMichaelJohnson: Improved readability and consistency with the JavaDoc in the other prebuilt transforms.  But more importantly, I then forgot to add another example using the params, which I have now done.

src/com/google/enterprise/adaptor/prebuilt/PrebuiltTransforms.java
BrettMichaelJohnson: removed short overloads

src/com/google/enterprise/adaptor/prebuilt/PrebuiltTransforms.java
wiarlawd: bump

src/com/google/enterprise/adaptor/prebuilt/PrebuiltTransforms.java
wiarlawd: Not buying readability in the source for a style guide violation.

src/com/google/enterprise/adaptor/prebuilt/PrebuiltTransforms.java
```diff
@@ -174,24 +228,53 @@ private CopyTransform(List<KeyPairing> copies, boolean overwrite,
...
-          log.log(Level.FINER, "Preexisting values for {0}. Combining: {1}",
-              new Object[] {kp.dest, destValues});
-          values.addAll(destValues);
+        switch (kp.dest.keyset) {
+          case METADATA:
+            Set<String> destValues = metadata.getAllValues(kp.dest.key);
+            if (!overwrite && !destValues.isEmpty()) {
+              log.log(Level.FINER, "Preexisting values for {0}. Combining: {1}",
+                  new Object[] {kp.dest, destValues});
+            values.addAll(destValues);
```
wiarlawd: Misindented.

test/com/google/enterprise/adaptor/prebuilt/PrebuiltTransformsTest.java
```diff
@@ -74,18 +74,238 @@ public void testCopy() {
...
+      golden.put("colour", "black");
+      golden.put("color", "black");
+      paramsGolden = Collections.unmodifiableMap(golden);
+    }
+    Metadata metadata = new Metadata();
+    Map<String, String> params = new HashMap<String, String>();
+    params.put("colour", "black");
+    transform.transform(metadata, params);
+    assertEquals(paramsGolden, params);
+  }
```
wiarlawd: Assert the metadata is still empty?

src/com/google/enterprise/adaptor/prebuilt/PrebuiltTransforms.java
```diff
@@ -174,24 +228,53 @@ private CopyTransform(List<KeyPairing> copies, boolean overwrite,
...
+            metadata.set(kp.dest.key, values);
+            break;
+          case PARAMS:
+            String value = values.iterator().next();
+            if (values.size() > 1) {
+              log.log(Level.FINER,
+                  "Multiple values for {0}. Using first value of {1} for {2}",
+                  new Object[] { kp.src, value, kp.dest });
+            }
+            params.put(kp.dest.key, value);
```
wiarlawd: You're not accounting for overwrite here.

test/com/google/enterprise/adaptor/prebuilt/PrebuiltTransformsTest.java
```diff
@@ -74,18 +74,238 @@ public void testCopy() {
...
+    config.put("1.to.keyset", "params");
+    config = Collections.unmodifiableMap(config);
+
+    MetadataTransform transform = PrebuiltTransforms.copyMetadata(config);
+
+    final Map<String, String> paramsGolden;
+    {
+      Map<String, String> golden = new HashMap<String, String>();
+      golden.put("colour", "black");
+      golden.put("color", "black");
```
wiarlawd, line 211: As discussed, without overwrite this should stay red.

test/com/google/enterprise/adaptor/prebuilt/PrebuiltTransformsTest.java
```diff
@@ -74,18 +74,238 @@ public void testCopy() {
...
+    config.put("1.to", "color");
+    config.put("1.to.keyset", "params");
     config = Collections.unmodifiableMap(config);
 
+    MetadataTransform transform = PrebuiltTransforms.copyMetadata(config);
+
+    final Map<String, String> paramsGolden;
+    {
+      Map<String, String> golden = new HashMap<String, String>();
+      golden.put("color", "black");
```
wiarlawd, line 336: Same here, should stay red without overwrite.

test/com/google/enterprise/adaptor/prebuilt/PrebuiltTransformsTest.java
```diff
@@ -118,6 +338,146 @@ public void testMove() {
   }
 
   @Test
+  public void testMoveFromParamsToMetadata() {
+    Map<String, String> config = new HashMap<String, String>();
+    config.put("overwrite", "true");
```
wiarlawd: Why are all the move tests specifying overwrite=true?

src/com/google/enterprise/adaptor/prebuilt/PrebuiltTransforms.java
```diff
@@ -335,48 +468,76 @@ public static MetadataTransform replaceMetadata(
...
+          new Object[] {key, toMatch, replacement, original});
+      Set<String> values = new HashSet<String>(original);
+      for (String value : original) {
+        String newValue = toMatch.matcher(value).replaceAll(replacement);
+        if (overwrite) {
+          values.remove(value);
         }
-        log.log(Level.FINE, "After replacing: {0}", values);
-        metadata.set(key, values);
+        values.add(newValue);
```
wiarlawd, line 528: I haven't thought deeply about it, but it seems like the set semantics used for values here could get us into trouble with overwrite. For example, suppose the original values are "w" and "ww", and the pattern replacement is "." to "$0$0" (the use of $ seems undocumented, as goes group 0; what am I missing?). Then "w" => "ww", and "w" is erased. Then "ww" => "www" and "ww" is erased. Now only "www" is left, but we started with two values.

src/com/google/enterprise/adaptor/prebuilt/PrebuiltTransforms.java
```diff
@@ -335,48 +468,76 @@ public static MetadataTransform replaceMetadata(
...
+      }
+      metadata.set(key, values);
+      log.log(Level.FINE, "After replacing metadata values for {0}: {1}",
+          new Object [] {key, values});
+    }
+
+    private void replaceInParams(String key, Map<String, String> params) {
+      String original = getTrimmedValue(params, key);
```
wiarlawd, line 536: I guess overwrite=false should just be a no-op here?

wiarlawd (changes requested):

src/com/google/enterprise/adaptor/prebuilt/PrebuiltTransforms.java
BrettMichaelJohnson: Done.

src/com/google/enterprise/adaptor/prebuilt/PrebuiltTransforms.java
BrettMichaelJohnson: Done.

src/com/google/enterprise/adaptor/prebuilt/PrebuiltTransforms.java
BrettMichaelJohnson: The JavaDoc says "If the destination is a param key, overwrite is ignored and its value is always replaced."

I might be convinced that overwrite=false makes a param final, if that would really be a desirable behavior.  The default value of overwrite is false, so making params write-once might be surprising behavior to the users.

src/com/google/enterprise/adaptor/prebuilt/PrebuiltTransforms.java
BrettMichaelJohnson: ibid

test/com/google/enterprise/adaptor/prebuilt/PrebuiltTransformsTest.java
BrettMichaelJohnson: Done.

test/com/google/enterprise/adaptor/prebuilt/PrebuiltTransformsTest.java
BrettMichaelJohnson: Copy/paste.

src/com/google/enterprise/adaptor/prebuilt/PrebuiltTransforms.java
```diff
@@ -174,24 +228,53 @@ private CopyTransform(List<KeyPairing> copies, boolean overwrite,
     @Override
     public void transform(Metadata metadata, Map<String, String> params) {
       for (KeyPairing kp : copies) {
-        Set<String> values = metadata.getAllValues(kp.src);
+        Set<String> values = new TreeSet<String>();
```
BrettMichaelJohnson: No longer need this TreeSet, since dropping support for metadata-or-params

wiarlawd:
While I'm thinking of it, s/prebuilt filters/prebuilt transforms/, or perhaps even better, PrebuiltTransforms.

src/com/google/enterprise/adaptor/prebuilt/PrebuiltTransforms.java
wiarlawd: Let's have a separate CL to fix this bug, by writing into a separate collection the way copy does.

src/com/google/enterprise/adaptor/prebuilt/PrebuiltTransforms.java
wiarlawd: OK, my preference is for the default value of overwrite to be false for metadata, but true for params. I think that produces the least surprising results, but still lets you explicitly set overwrite=false if you want to not change a previously set value.

I was thinking that the constructor argument and field could be a Boolean: TRUE, FALSE, or null, where null is interpreted as FALSE by dest.keyset == METADATA, and TRUE by dest.keyset == PARAMS.

test/com/google/enterprise/adaptor/prebuilt/PrebuiltTransformsTest.java
```diff
@@ -478,13 +546,69 @@ public void testMoveFromParamToParam() {
...
+    config.put("1.to.keyset", "params");
+    config = Collections.unmodifiableMap(config);
+
+    MetadataTransform transform = PrebuiltTransforms.moveMetadata(config);
+
+    final Metadata metadataGolden = new Metadata().unmodifiableView();
+    final Map<String, String> paramsGolden;
+    {
+      Map<String, String> golden = new HashMap<String, String>();
+      golden.put("color", "black");
```
wiarlawd, line 592: I was surprised that colour was gone in this case, but I think it's just a random choice, and we don't want Yet Another Parameter for this. Should we include a note in the moveTransform metadata that if the destination is a param and overwrite is false, the source is still removed?

wiarlawd (approved):

test/com/google/enterprise/adaptor/prebuilt/PrebuiltTransformsTest.java
BrettMichaelJohnson: I was wrestling with this. I *could* try to avoid the delete if the move was a no-op, but overwrite == false for params is just messed up enough that I am not sure it is worth the effort.

PapaJoltOmega (approved):
Thank you

PR #43, opened 2017-05-06
Replace Corpora with Keyset - Part1

BrettMichaelJohnson:


src/com/google/enterprise/adaptor/MetadataTransform.java
```diff
@@ -25,6 +25,7 @@
  */
 public interface MetadataTransform {
 
+  /** Keys for data stored in the {@code Map} of {@code params}. */
```
wiarlawd: Change to a // or /* comment, since this would be completely misleading to read in javadoc.

src/com/google/enterprise/adaptor/MetadataTransform.java
```diff
@@ -35,6 +36,41 @@
...
 
+  /**
+   * Identifies the location a configured key/value pair resides,
+   * either in the {@link Metadata}, or the map of {@code params}.
+   * <p>
+   * Suppose a transform wishes examine or modify the display URL, which is
+   * stored in the {@code params}. The transform's configuration might look
+   * like:
+   * <pre><code>
+     key=Display-URL
```
wiarlawd: Leaving out the *'s is unnecessary and possibly incorrect.

src/com/google/enterprise/adaptor/MetadataTransform.java
```diff
@@ -35,6 +36,41 @@
...
+    METADATA("metadata"),
+    PARAMS("params");
+
+    private final String name;
+
+    private Keyset(String name) {
+      this.name = name;
+    }
+
+    public static Keyset from(String val) {
```
wiarlawd: Add javadoc and a throws tag for IllegalArgumentException, and do the same for Transmission-Decision. The enum values could stand some javadoc, too. There is no mention in the javadoc of what the string names for the enum values actually are. Don't forget that you can add javadoc to the individual values of the enum as well (e.g., METADATA and PARAMS).

src/com/google/enterprise/adaptor/MetadataTransform.java
```diff
@@ -35,6 +36,41 @@
   public static final String KEY_TRANSMISSION_DECISION 
       = "Transmission-Decision";
 
+  /**
+   * Identifies the location a configured key/value pair resides,
+   * either in the {@link Metadata}, or the map of {@code params}.
+   * <p>
+   * Suppose a transform wishes examine or modify the display URL, which is
```
wiarlawd: s/wishes/wishes to/

src/com/google/enterprise/adaptor/MetadataTransform.java
```diff
@@ -35,6 +36,41 @@
...
+   * stored in the {@code params}. The transform's configuration might look
```
wiarlawd: s/look like/include/

src/com/google/enterprise/adaptor/MetadataTransform.java
```diff
@@ -35,6 +36,41 @@
...
+    PARAMS("params");
+
+    private final String name;
+
+    private Keyset(String name) {
+      this.name = name;
+    }
+
+    public static Keyset from(String val) {
+      return (val == null) ? METADATA : Keyset.valueOf(val.toUpperCase());
```
wiarlawd: Add US locale.

src/com/google/enterprise/adaptor/MetadataTransform.java
```diff
@@ -54,6 +90,7 @@ public static TransmissionDecision from(String val) {
       return TransmissionDecision.valueOf(val.replace('-', '_').toUpperCase());
     }
 
+    @Override
```
wiarlawd: Add javadoc for these, too, just to say that they return the names.

wiarlawd (commented):
Wordsmithing and java doc.

John L

src/com/google/enterprise/adaptor/MetadataTransform.java
BrettMichaelJohnson: Added JavaDoc for each key.

src/com/google/enterprise/adaptor/MetadataTransform.java
BrettMichaelJohnson: Done.

src/com/google/enterprise/adaptor/MetadataTransform.java
BrettMichaelJohnson: include

src/com/google/enterprise/adaptor/MetadataTransform.java
```diff
@@ -25,34 +27,57 @@
...
-   * stored in the {@code params}. The transform's configuration might look
-   * like:
+   * Suppose a transform wishes to examine or modify the display URL, which is
+   * stored in the {@code params}. The transform's configuration might include:
    * <pre><code>
      key=Display-URL
      keyset=params
      </code></pre>
-   * If unspecified, the default value is {@code metadata}.
+   * If unspecified, the default value is {@code METADATA}.
```
wiarlawd: I don't like this change, since in this context, it's the name that matters, not the enum identifier.

src/com/google/enterprise/adaptor/MetadataTransform.java
```diff
@@ -25,34 +27,57 @@
...
   public static final String KEY_CONTENT_TYPE = "Content-Type";
+
+  /** The key for the document's crawl once flag supplied in the {@code params}.
+   */
   public static final String KEY_CRAWL_ONCE = "Crawl-Once";
+
+  /** The key for the document's lock flag supplied in the {@code params}. */
   public static final String KEY_LOCK = "Lock";
+
+  /** The key for the document's last modified timestamp supplied in the
```
wiarlawd: This is a style guide violation. Implementation comments can be formatted like this, but not javadoc.

https://google.github.io/styleguide/javaguide.html#s7.1-javadoc-formatting

src/com/google/enterprise/adaptor/MetadataTransform.java
```diff
@@ -61,20 +86,43 @@ private Keyset(String name) {
...
+  /**
+   * Transforms can cancel sending a document, or cancel sending its contents by
+   * adding a {@code Transmission-Decision} to the {@code Map} of {@code params}
+   * supplied to the {@code transform} method of the {@code MetadataTransform}.
+   * If no {@code Transmission-Decision} is set in the params, the default value
+   * of {@code AS_IS} is used.
+   */
   public enum TransmissionDecision {
+    /** The document's metadata and content are to remain intact.
+     *  This value's name is {code as-is}.
```
wiarlawd: Missing @'s on these.

wiarlawd (changes requested):

src/com/google/enterprise/adaptor/MetadataTransform.java
BrettMichaelJohnson: Done.

src/com/google/enterprise/adaptor/MetadataTransform.java
BrettMichaelJohnson: Done.

PapaJoltOmega (commented):
Is the plan to allow skip document filter to use name corpora and name keyset or just one?

Why not make a CL that changes how SkipDocumentFilter works, rather than this detached structure?

Thank you

wiarlawd:
Brett and I have been talking about SkipDocumentFilter. Since we dropped support for "metadata or params", it can't be rewritten as a small wrapper on RegexFilter, which was designed to replace it. We weren't sure whether it made sense to leave it alone (and be certain of not breaking something) or add keyset as a small nod to consistency. None of the existing pull requests are addressing SkipDocumentFilter and corpora vs keyset.

This change is actually part of adding Keyset to the other transforms. Temporally, this isn't part 1 at all, but part 2:

Part 1: Add keyset to RegexFilter (in place of corpora), DateFilter (in place of corpora), and PrebuiltTransforms (new)
Part 2: Add a shared Keyset enum to MetadataTransform (this CL)
Part 3: Edit the transforms to use the shared Keyset.

It's a funny order, but most of the coding in part 1 had been finished before the renaming decision was made, so parts 2 and 3 were not part of the original plan, and the filters all had their own nested Corpora enum originally.

src/com/google/enterprise/adaptor/MetadataTransform.java
```diff
@@ -25,20 +27,107 @@
...
+   * either in the {@link Metadata}, or the map of {@code params}
+   * supplied to the {@link #transform} method of a {@code MetadataTransform}.
+   * <p>
+   * Suppose a transform wishes to examine or modify the display URL, which is
+   * stored in the {@code params}. The transform's configuration might include:
+   * <pre><code>
+     key=Display-URL
+     keyset=params
+     </code></pre>
+   * If unspecified, the default value is {@code METADATA}.
```
wiarlawd: Missed this? s/METADATA/metadata/

src/com/google/enterprise/adaptor/MetadataTransform.java
```diff
@@ -25,20 +27,107 @@
...
-  /** Transforms can cancel sending doc, or cancel sending its contents. */
+  /**
+   * Identifies the location a configured key/value pair resides,
+   * either in the {@link Metadata}, or the map of {@code params}
+   * supplied to the {@link #transform} method of a {@code MetadataTransform}.
+   * <p>
+   * Suppose a transform wishes to examine or modify the display URL, which is
+   * stored in the {@code params}. The transform's configuration might include:
+   * <pre><code>
+     key=Display-URL
```
wiarlawd: Add *'s

src/com/google/enterprise/adaptor/MetadataTransform.java
```diff
@@ -25,20 +27,107 @@
...
+     keyset=params
+     </code></pre>
+   * If unspecified, the default value is {@code METADATA}.
+   *
+   * @since 4.1.4
+   */
+  public enum Keyset {
+    /** The key/value(s) reside in the {@link Metadata}.
```
wiarlawd: Add line breaks for javadoc formatting in these two.

wiarlawd (changes requested):
Nits

src/com/google/enterprise/adaptor/MetadataTransform.java
BrettMichaelJohnson: Done.

src/com/google/enterprise/adaptor/MetadataTransform.java
```diff
@@ -25,20 +27,107 @@
...
+   * <p>
+   * Suppose a transform wishes to examine or modify the display URL, which is
+   * stored in the {@code params}. The transform's configuration might include:
+   * <pre><code>
+     key=Display-URL
+     keyset=params
+     </code></pre>
+   * If unspecified, the default value is {@code METADATA}.
+   *
+   * @since 4.1.4
```
BrettMichaelJohnson: Add // @since TODO(bmj)

BrettMichaelJohnson: Done.

src/com/google/enterprise/adaptor/MetadataTransform.java
BrettMichaelJohnson: Done.

src/com/google/enterprise/adaptor/MetadataTransform.java
BrettMichaelJohnson: I thought this over, and METADATA is correct as the value of the enum.

src/com/google/enterprise/adaptor/MetadataTransform.java
```diff
@@ -65,20 +66,23 @@
...
-     keyset=params
-     </code></pre>
+   * key=Display-URL
+   * keyset=params
+   * </code></pre>
    * If unspecified, the default value is {@code METADATA}.
    *
    * @since 4.1.4
    */
+  // @since TODO(bmj)
```
wiarlawd: Belay that order.

src/com/google/enterprise/adaptor/MetadataTransform.java
```diff
@@ -25,20 +27,111 @@
...
+      this.name = name;
+    }
+
+    /**
+     * Returns a {@code Keyset} value with the given {@code name}, or the
+     * default value of {@code METADATA} if name is {@code null}.
+     * @param name the name of the Keyset value
+     * @throws IllegalArgumentException if there is no Keyset value with
+     *         {@code name}.
+     */
```
wiarlawd: javadoc is warning about a missing return tag.

src/com/google/enterprise/adaptor/MetadataTransform.java
```diff
@@ -47,13 +140,23 @@ private TransmissionDecision(String n) {
...
 
-    public static TransmissionDecision from(String val) {
-      if (null == val) {
+    /**
+     * Returns a {@code TransmissionDecision} value with the given {@code name},
+     * or the default value of {@code AS_IS} if name is {@code null}.
+     * @param name the name of the TransmissionDecision value
+     * @throws IllegalArgumentException if there is no TransmissionDecision
+     *         value with {@code name}.
+     */
```
wiarlawd: javadoc is warning about a missing return tag.

src/com/google/enterprise/adaptor/MetadataTransform.java
```diff
@@ -25,20 +27,111 @@
...
+   * {@code params}.
+   */
   public static final String KEY_TRANSMISSION_DECISION 
       = "Transmission-Decision";
 
-  /** Transforms can cancel sending doc, or cancel sending its contents. */
+  /**
+   * Identifies the location a configured key/value pair resides,
+   * either in the {@link Metadata}, or the map of {@code params}
+   * supplied to the {@link #transform} method of a {@code MetadataTransform}.
```
wiarlawd: link here, but code for TransmissionDecision. Also, most of the time for a method link, you want to specify the text: s/#transform/#transform transform/

src/com/google/enterprise/adaptor/MetadataTransform.java
wiarlawd: I'll buy that.

wiarlawd (changes requested):
Sorry, I was looking at the javadoc over the metadata vs METADATA issue.

PapaJoltOmega:
> been talking about SkipDocumentFilter. Since we dropped 
> support for "metadata or params", it can't be rewritten as a 
> small wrapper on RegexFilter, which was designed to replace 
> it. We weren't sure whether it made sense to leave it alone 
> (and be certain of not breaking something) or add keyset as 
> a small nod to consistency. None of the existing pull requests 
> are addressing SkipDocumentFilter and corpora vs keyset.

What do you want to do? 

Do you want to change SkipDocumentFilter in backwards incompatible way?

or do we want to add metadata_and_params while making default be metadata?

I would like if we would revist this CL after checking in some of the items that will use it.  That way the code is used when it is submitted. 

Thank you

wiarlawd:
I don't have a super-strong preference for whether to add keyset to SkipDocumentFilter. I would lean slightly toward adding it for consistency. That would be a separate change. We do not want to break backward compatibility of existing transform configs, or add METADATA_OR_PARAMS to Keyset (because it's prone to hard to understand failures and doesn't add new functionality, beyond slightly simplifying transforms on the params by not having to specify the keyset).

So what you want here is to combine this change with Brett's planned part 3, to introduce Keyset in MetadataTransform and remove it from the individual transforms in one change? Seems reasonable to me either way. Brett was just trying to create smaller, incremental changes.

PapaJoltOmega:
(1) Do we want to deprecate SkipDocumentFilter?

(2) I would prefer to read the combined part 2+3.

Instead of breaking CLs into:
CL 1: make pieces
CL 2: use pieces

I think there are advantages to:
CL 1: make part of the pieces and use those pieces
CL 2: make another part of the pieces and use those pieces

(Let me know if that's not clear)

http://googlegsa.github.io/adaptor/docs/DevelopmentProperties.html -- It's the preference for vertical slices and eschewing horizontal slices. Even here where I thinK i know where we are going I would like to combine part 2+3, please.

Thank you very much

src/com/google/enterprise/adaptor/MetadataTransform.java
BrettMichaelJohnson: Done.

src/com/google/enterprise/adaptor/MetadataTransform.java
BrettMichaelJohnson: Done.

src/com/google/enterprise/adaptor/MetadataTransform.java
BrettMichaelJohnson: Done.

BrettMichaelJohnson:
PJ, I would rather have done this as one CL, but split it up to avoid merge conflicts with all the other outstanding CLs.

PapaJoltOmega (approved):
Makes sense to avoid merge.  Thank you.  LGTM.

src/com/google/enterprise/adaptor/MetadataTransform.java
```diff
@@ -25,20 +27,114 @@
...
+   * stored in the {@code params}. The transform's configuration might include:
+   * <pre><code>
+   * key=Display-URL
+   * keyset=params
+   * </code></pre>
+   * If unspecified, the default value is {@code METADATA}.
+   *
+   * @since 4.1.4
+   */
+  // @since TODO(bmj)
```
wiarlawd: You can remove this TODO.

wiarlawd (approved):
LGTM, with a nit.

src/com/google/enterprise/adaptor/MetadataTransform.java
BrettMichaelJohnson: Done.

PR #45, opened 2017-05-07
Move addParam to Response2, a new subinterface of Response.

wiarlawd:
This preserves source and binary compatibility when adding new methods
to the interfaces. In the future, new methods can be added to the
Response2 interface, rather than creating additional subinterfaces.

When calling an adaptor, an instance of Response2 is always provided.

Adaptor implementers who want to use the methods of Response2 should
cast the Response parameter passed to getDocContent to Response2, and
for unit tests, should extend the new UnsupportedResponse class,
rather than implementing Response2 directly.


BrettMichaelJohnson (approved):
LGTM

wiarlawd:
The current plan is to not introduce the DocIdPusher and Response subinterfaces.

PR #46, opened 2017-05-08
Add a caching MockResponse class for unit testing.

wiarlawd:
This class is a copy of WrapperAdaptor.GetContentsResponse. That class
was also copied to the SharePoint connector, and very similar classes
written for the file system, Documentum, and OpenText connectors. The
library and SharePoint versions were the most complete; the others
were missing a handful of the flag methods (e.g., setNoFollow).

Only minor changes were made, including some merging of the way the
output stream, metadata, and anchors were handled in the various
implementations. The fields and methods were rearranged to match the
Response source file.


BrettMichaelJohnson (approved):
LGTM

src/com/google/enterprise/adaptor/testing/MockResponse.java
```diff
@@ -0,0 +1,221 @@
...
+public class MockResponse implements Response {
+  private final OutputStream os;
+  private boolean notModified;
+  private boolean notFound;
+  private boolean noContent;
+  private String contentType;
+  private Date lastModified;
+  private Metadata metadata = new Metadata();
+  private Acl acl;
+  private Map<String, Acl> namedResources = new HashMap<String, Acl>();
```
PapaJoltOmega: How about TreeMap

PapaJoltOmega (commented):
Is it fitting to look at this CL as a rename of a class and a move of it to testing? Thank you

src/com/google/enterprise/adaptor/testing/MockResponse.java
wiarlawd: It's unnecessary. Set and Map equality are already order-independent, and equality is all the tests care about here, where they have the map itself to hand. On the other hand, if we were testing from other end of the wire (e.g., the GSA), then this map would get stringified, and we would need to care about the iteration order in order to (easily) check equality.

wiarlawd (commented):
Yes, that's a fair characterization. It started out as a mashup of classes from the connectors, before I noticed that the library had essentially the same class in WrapperAdaptor. I can update the description to s/Add/Expose/ and be more clear about making the class public in the testing package in the first sentence of the description.

wiarlawd:
This class is really a fake and not a mock, and there's an AccumulatingDocIdPusher as something of a peer. I think we should name them similarly in the testing package. I'm thinking of RecordingResponse and RecordingDocIdPusher. Fake is too generic, Caching seems like it adds a cache to something else, and Accumulating seems too specific to DocIdPusher, where you can keep pouring in data with the same methods. Any preferences or suggestions?

src/com/google/enterprise/adaptor/testing/RecordingResponse.java
```diff
@@ -0,0 +1,222 @@
...
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * A fake implementation of {@link Response} that simply records the
+ * values it receives.
+ */
+public class RecordingResponse implements Response {
+  // TODO(jlacey): Implement Response2.
```
PapaJoltOmega: You've convinced me that we can require changes to src and that we will sufficiently advertise and assist with those changes. So I think we are both saying no Response2, now.

PapaJoltOmega (approved):
Thank you.

Looks like a fine transformation. Is there anything in particular that I should take a look at?

wiarlawd:
The crux of the change is replacing multiple similar implementations in the library and connectors with a single shared implementation in a new testing package, as part of the src tree and adaptor.jar. The connectors will be updated (post-4.1.3) to use this shared class.

PapaJoltOmega (approved):
LGTM

PR #47, opened 2017-05-08
Mock the call order semantics of the Response interface.

wiarlawd:
Other changes:

* Fix the CommandStreamParser and CommandStreamParserTest, which
  violated these constraints.


BrettMichaelJohnson (approved):
LGTM

test/com/google/enterprise/adaptor/CommandStreamParserTest.java
```diff
@@ -167,7 +167,6 @@ public void testSupportedDelimiterCharacters() throws Exception {
   public void testRetriever() throws IOException {
     String source = "GSA Adaptor Data Version 1 [\n]\n"
         + "id=123\n"
-        + "up-to-date\n"
```
PapaJoltOmega, line 170: why does this line get removed?

src/com/google/enterprise/adaptor/CommandStreamParser.java
```diff
@@ -467,6 +467,19 @@ public void readFromRetriever(DocId docId, Response response) throws IOException
...
+          // Finish by putting accumulated ACL into response.
+          if (sendAclWithDocument) {
+            aclBuilder.setPermits(permits);
+            aclBuilder.setDenies(denies);
+            response.setAcl(aclBuilder.build());
+          }
+          break;
+        default:
+      }
+      switch (command.getOperation()) {
```
PapaJoltOmega: so instead of 1 command at a time we are doing 2 commands at a time?

PapaJoltOmega (commented):
This CL is cool.  It looks like it adds rigid invariant checking. A couple quick questions, please.  Thank you

src/com/google/enterprise/adaptor/CommandStreamParser.java
wiarlawd: Sort of. There's no "end of ACL" tag in the input, so the code was waiting until the end of the input (that is, the command loop) to process the ACL. But it's too late to call setAcl on the Response at that point, because respondXXX or getOutputStream have already been called. So I moved this block to the top of the loop, and check to see if we're about to process one of the must-be-last methods, and use that as a trigger to make sure that setAcl is called while it still can be.

test/com/google/enterprise/adaptor/CommandStreamParserTest.java
wiarlawd: There's content buried at the end of this input string, and you can't have both (since they correspond to calling respondNotModified and getOutputStream). I picked the content here to preserve the other things this test was checking, and added a separate test for up-to-date, and a test for the two commands conflicting, and one more for trying to do something else you can't do at the end (similar to setAcl).

src/com/google/enterprise/adaptor/CommandStreamParser.java
PapaJoltOmega: Ah, I was reading this code wrong.  In the CL review, I was thinking somehow that getOperation() also does a read.  Which is completely the wrong reading.  

Can we do command.getOperation() once and use that gotten value in the two switches?

Something about the switch made me think it was always processing new token and that somehow we were processing two token at a time.  Which is completely the wrong reading, admitadly.

PapaJoltOmega (approved):
Great.  Thank you.  LGTM.

src/com/google/enterprise/adaptor/CommandStreamParser.java
wiarlawd: I pondered doing so initially, so I was tempted to say sure. However, this class calls command.getOperation() multiple times after readCommand() in four separate places. I changed the comment a little bit for clarity. Another option would be put this code into a setAcl or finishAcl helper method, and instead of this extra switch statement, call the ACL helper separately in the corresponding cases in the existing switch statement.

PapaJoltOmega (approved):
LGTM.,  Thank you

PapaJoltOmega (approved):
LGTM

PR #48, opened 2017-05-09
Dissolved most of WrapperAdaptor.

wiarlawd:
Most of WrapperAdaptor was unused. Its reason for existence was
AutoUnzip, which was deleted 4 years ago in commit 85c51d5.

WrapperAdaptor: Used trivially by StatRpcMethodTest, replaced by Proxy
WrapperRequest: Unused
WrapperResponse: Unused
GetContentsRequest: Unused
GetContentsResponse: Moved to the testing package in #46
WrapperDocIdPusher: Unused
WrapperAdaptorContext: Used trivially by TestHelper and 3 connectors

WrapperAdaptorContext is left here for backward compatibility
with existing connectors, but may be replaced later.


test/com/google/enterprise/adaptor/TestHelper.java
```diff
@@ -87,17 +89,6 @@ public void setAuthzAuthority(AuthzAuthority authzAuthority) {}
     return getDocIds(adaptor, Collections.<String, String>emptyMap());
   }
 
-  public static byte[] getDocContent(Adaptor adaptor, DocId docId)
```
BrettMichaelJohnson, line 90: TestHelper is public and so is this method, so they are both available to connectors. I don't think you should remove this method.

BrettMichaelJohnson (changes requested):

test/com/google/enterprise/adaptor/TestHelper.java
```diff
@@ -118,4 +109,17 @@ public static void initSSLKeystores() {
         TestHelper.class.getResource("/test-cacerts.jks").getPath());
     System.setProperty("javax.net.ssl.trustStorePassword", "changeit");
   }
+
+  /** Gets a proxy for the given class that does nothing. */
+  public static <T> T newProxyInstance(Class<T> clazz) {
```
PapaJoltOmega: how about:
   s/newProxyInstance/doNothingMock/

test/com/google/enterprise/adaptor/TestHelper.java
```diff
@@ -118,4 +109,17 @@ public static void initSSLKeystores() {
         TestHelper.class.getResource("/test-cacerts.jks").getPath());
     System.setProperty("javax.net.ssl.trustStorePassword", "changeit");
   }
+
+  /** Gets a proxy for the given class that does nothing. */
```
PapaJoltOmega: s/that does nothing/where every method is a no-op/

PapaJoltOmega (commented):
Thank you

test/com/google/enterprise/adaptor/TestHelper.java
wiarlawd: Would you accept doNothingProxy? I'd like to keep Proxy in the names of things that use Proxy.

test/com/google/enterprise/adaptor/TestHelper.java
wiarlawd: None of our connectors use the test classes from the library. They all reference the library through adaptor-withlib.jar. Four connectors have copies of TestHelper, but none of them use this one, or anything out of WrapperAdaptor except for WrapperAdaptorContext as noted. If third-party connectors are referring to things in the library tests, I have no sympathy.

PapaJoltOmega: Thanks for the consideration.  

Agree with John that reaching into test subdir is tacit agreement to higher level of disruption.

PapaJoltOmega (approved):
Thanks for ameliorating code debt. LGTM

PapaJoltOmega (approved):
LGTM

BrettMichaelJohnson (approved):
LGTM

PR #49, opened 2017-05-11
Fix b/21168971: Fix NoColor nested class ref

BrettMichaelJohnson:


PapaJoltOmega (commented):
Does this CL mean we should make RC3 of library?

wiarlawd:
No, this logging.properties only affects library developers, so I don't think it's worth another RC. It's the logging.properties in the connector installers that matters for customers. (Whether the split is a useful one is debatable, and we are making them more similar to each other, with the exception that the file handler is disabled in the GitHub versions.)

PapaJoltOmega (approved):

BrettMichaelJohnson:
Also note that the error is in a commented out section of this logging.properties, so even day-to-day library developers are unlikely to encounter it.

PR #51, opened 2017-05-16
Add UnsupportedOperationException mocks for Adaptor args.

wiarlawd:
Add basic mocks, whose methods always throw an exception, for the
interfaces that are used as a parameter in an Adaptor interface
method: AdaptorContext, DocIdPusher, Request, and Response.


src/com/google/enterprise/adaptor/testing/UnsupportedDocIdPusher.java
```diff
@@ -0,0 +1,103 @@
...
+  public DocId pushNamedResources(Map<DocId, Acl> resources)
+      throws InterruptedException {
+    throw new UnsupportedOperationException(
+        "UnsupportedDocIdPusher was called");
+  }
+
+  /** @throws UnsupportedOperationException always */
+  @Override
+  public DocId pushNamedResources(Map<DocId, Acl> resources,
+                                  ExceptionHandler handler)
```
BrettMichaelJohnson: Nit. I would rewrap these lines.

src/com/google/enterprise/adaptor/testing/UnsupportedAdaptorContext.java
```diff
@@ -0,0 +1,131 @@
...
+ * {@code UnsupportedOperationException} if any method is called.
+ *
+ * <p>This class is intended to be extended for unit testing, rather
+ * than implementing the {@link AdaptorContext} interface directly.
+ */
+public class UnsupportedAdaptorContext implements AdaptorContext {
+  /** @throws UnsupportedOperationException always */
+  @Override
+  public Config getConfig() {
+    throw new UnsupportedOperationException("UnsupportedResponse was called");
```
BrettMichaelJohnson: s/UnsupportedResponse/UnsupportedAdaptorContext/g

BrettMichaelJohnson (changes requested):

BrettMichaelJohnson (approved):
LGTM

PapaJoltOmega (commented):
The goal here is to future proof interfaces for additions?  And to avoid having developers change their code?  

Yes?

Thank you

src/com/google/enterprise/adaptor/AdaptorContext.java
```diff
@@ -14,13 +14,20 @@
...
 import com.sun.net.httpserver.HttpContext;
 import com.sun.net.httpserver.HttpExchange;
 import com.sun.net.httpserver.HttpHandler;
 
 /**
  * Methods for an Adaptor to communicate with the adaptor library.
  * Implementations of this class must be thread-safe.
+ *
+ * <p>This interface should not be implemented directly for unit testing.
+ * Instead, use a mocking framework, such as Mockito, a
```
PapaJoltOmega: How about this alternate text with `because' and lower MOckito:

Avoid implementing this interface in adaptor unit tests because new methods may
be added in the future. Instead use {@link UnsupportedAdaptorContext}, or use
an automated mock generator like Mockito or {@code java.lang.reflect.Proxy}.

PapaJoltOmega (approved):
Thank you.  LGTM

wiarlawd:
I'm going to submit this. I still want to think about other alternatives before the 4.1.4 release, but I'd like this to be the default instead of doing nothing.

PR #52, opened 2017-05-18
Fix b/38357564 Null Param value throws NPE from Filters

BrettMichaelJohnson:


src/com/google/enterprise/adaptor/prebuilt/RegexFilter.java
```diff
@@ -204,7 +204,12 @@ private boolean foundInMetadata(Metadata metadata) {
   private boolean foundInParams(Map<String, String> params) {
     boolean found = false;
     if (params.containsKey(key)) {
-      found = pattern.matcher(params.get(key)).find();
+      String value = params.get(key);
+      if (value == null) {
+        found = "\\A".equals(pattern.toString());
```
wiarlawd: What do you think about value = "" instead, so other patterns matching empty values would also match, and to avoid treating \A as magic?

wiarlawd (approved):

PapaJoltOmega (approved):
LGTM

PR #53, opened 2017-05-18
Improve log messages in SkipDocumentFilter and RegexFilter

BrettMichaelJohnson:
Fix b/38359039 SkipDocumentFilter should log "No transmission decision"
instead of "Not Skipping"

Fix b/38356739 Param and metadata names are case-sensitive in transforms

### DIFF
--- a/src/com/google/enterprise/adaptor/prebuilt/RegexFilter.java
+++ b/src/com/google/enterprise/adaptor/prebuilt/RegexFilter.java
@@ -21,6 +21,7 @@ import com.google.enterprise.adaptor.Metadata;
 import com.google.enterprise.adaptor.MetadataTransform;
 
 import java.util.Map;
+import java.util.Set;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import java.util.regex.Pattern;
@@ -185,7 +186,15 @@ public class RegexFilter implements MetadataTransform {
    */
   private boolean foundInMetadata(Metadata metadata) {
     boolean found = false;
-    for (String value : metadata.getAllValues(key)) {
+    Set<String> values = metadata.getAllValues(key);
+    if (values.isEmpty()) {
+      if (metadata.getKeys().contains(key)) {
+        log.finest("No values for key `" + key + "' in metadata.");
+      } else {
+        log.finest("No key `" + key + "' in metadata.");
+      }
+    }
+    for (String value : values) {
       if (pattern.matcher(value).find()) {
         found = true;
         break;
@@ -205,6 +214,8 @@ public class RegexFilter implements MetadataTransform {
     boolean found = false;
     if (params.containsKey(key)) {
       found = pattern.matcher(params.get(key)).find();
+    } else {
+      log.finest("No key `" + key + "' in params.");
     }
     log.fine((found ? "Did" : "Did not") + " find matching pattern for key `"
         + key + "' in params.");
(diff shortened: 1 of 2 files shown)


PapaJoltOmega (approved):
LGTM

src/com/google/enterprise/adaptor/prebuilt/RegexFilter.java
```diff
@@ -185,7 +186,15 @@ private RegexFilter(String key, Keyset keyset, Pattern pattern,
    */
   private boolean foundInMetadata(Metadata metadata) {
     boolean found = false;
-    for (String value : metadata.getAllValues(key)) {
+    Set<String> values = metadata.getAllValues(key);
+    if (values.isEmpty()) {
+      if (metadata.getKeys().contains(key)) {
+        log.finest("No values for key `" + key + "' in metadata.");
```
wiarlawd: Include docid in these messages, and use MessageFormat style.

wiarlawd (changes requested):
This is not a fix for b/38356739, it just adds some logging suggested in that bug. I would move the other bug into the summary and add these an an "other change" in the description.

src/com/google/enterprise/adaptor/prebuilt/SkipDocumentFilter.java
```diff
@@ -142,7 +144,15 @@ private SkipDocumentFilter(String propertyName, Pattern pattern,
    */
   private boolean foundInMetadata(Metadata metadata) {
     boolean found = false;
-    for (String value : metadata.getAllValues(propertyName)) {
+    Set<String> values = metadata.getAllValues(propertyName);
+    if (values.isEmpty() && log.isLoggable(Level.FINEST)) {
+      if (metadata.getKeys().contains(propertyName)) {
+        log.finest("No values for key `" + propertyName + "' in metadata.");
```
wiarlawd: The use of ` feels like an affectation, and is used only 8 times in the full googlegsa code base (compared to 141 occurrences of '"). Let's not add 6 more. If we takeout the existing two each in SkipDocumentFilter and RegexDocumentFilter, then the library would have zero.

src/com/google/enterprise/adaptor/prebuilt/SkipDocumentFilter.java
```diff
@@ -162,6 +172,8 @@ private boolean foundInParams(Map<String, String> params) {
     boolean found = false;
     if (params.containsKey(propertyName)) {
       found = pattern.matcher(params.get(propertyName)).find();
+    } else if (log.isLoggable(Level.FINEST)) {
```
wiarlawd: Let's use MessageFormat-style on this one.

src/com/google/enterprise/adaptor/prebuilt/RegexFilter.java
```diff
@@ -205,9 +214,11 @@ private boolean foundInParams(Map<String, String> params) {
     boolean found = false;
     if (params.containsKey(key)) {
       found = pattern.matcher(params.get(key)).find();
+    } else if (log.isLoggable(Level.FINEST)) {
```
wiarlawd: The isLoggable check is redundant here.

wiarlawd (approved):
LGTM, with a nit.

src/com/google/enterprise/adaptor/prebuilt/RegexFilter.java

PR #54, opened 2017-05-20
Expose a recording DocIdPusher fake for unit testing.

wiarlawd:
Part 1: Merge the various implementations.

Rename AccumulatingDocIdPusher to RecordingDocIdPusher, make it
public, and move it to the testing package.This class was also copied
to five of the connectors, where they slightly diverged. The
differences were merged into this copy. Only the library's version
extended AbstractDocIdPusher, but they all contained exact copies of
AbstractDocIdPusher's methods. The methods were rearranged for the
getters to follow the setters.

Semantic changes and javadoc will follow in later changes. This one
contains the necessary functional changes, and comes in just over
git's default similarity index of 50% for recognizing renamed files.


src/com/google/enterprise/adaptor/testing/RecordingDocIdPusher.java
```diff
@@ -12,57 +12,92 @@
...
       Map<GroupPrincipal, ? extends Collection<Principal>> defs,
       boolean caseSensitive, ExceptionHandler handler)
       throws InterruptedException {
-    groups.putAll(defs); 
+    for (GroupPrincipal key : defs.keySet()) {
+      Collection<Principal> members = defs.get(key);
+      if (members instanceof List) {
+        groups.put(key, unmodifiableList(new ArrayList<Principal>(members)));
+      } else if (members instanceof java.util.Set) {
+        groups.put(key, unmodifiableSet(new LinkedHashSet<Principal>(members)));
```
BrettMichaelJohnson: I'm assuming #include java.util.Set put you over the %50 diffs

BrettMichaelJohnson (approved):

src/com/google/enterprise/adaptor/testing/RecordingDocIdPusher.java
wiarlawd: No, just forgotten.

src/com/google/enterprise/adaptor/testing/RecordingDocIdPusher.java
```diff
@@ -12,57 +12,93 @@
...
+
+  public Map<DocId, Acl> getNamedResources() {
+    return Collections.unmodifiableMap(namedResources);
+  }
+
+  public Map<GroupPrincipal, Collection<Principal>> getGroupDefinitions() {
+    return Collections.unmodifiableMap(groups);
+  }
+
+  public void reset() {
```
PapaJoltOmega, line 98: why reset instead of make new one?

PapaJoltOmega: i see that it is an existing method. the source moved and i didn't see method already existed.

src/com/google/enterprise/adaptor/testing/RecordingDocIdPusher.java
```diff
@@ -12,57 +12,93 @@
...
 
   @Override
   public GroupPrincipal pushGroupDefinitions(
       Map<GroupPrincipal, ? extends Collection<Principal>> defs,
       boolean caseSensitive, ExceptionHandler handler)
       throws InterruptedException {
-    groups.putAll(defs); 
+    for (GroupPrincipal key : defs.keySet()) {
+      Collection<Principal> members = defs.get(key);
+      if (members instanceof List) {
```
PapaJoltOmega, line 71: what does the extra treatment of List and Set gain?

PapaJoltOmega (approved):
Thank you

src/com/google/enterprise/adaptor/testing/RecordingDocIdPusher.java
wiarlawd: Most of the tests in the connectors use assertEquals on the collected groups. Most collections are not equal to each other, even if they contain the same members, because they use Object.equals. List and Set are different, but a List only equals another List, and a Set only equals another Set. Some connectors use a Set internally, and others a List. The tests have to use the same type. This code is trying to make a defensive copy, based on Eric's copy in the SharePoint connector. But you can't make a defensive copy that preserves order and equality without knowing whether you are copying a List or a Set.

It is possible that a third-party connector would use another collection type that supports membership equality, but that's rather unlikely, and I just punted on that. (Specifically, Deque and Queue, the other JDK collections, do not support membership equality. Guava has some collections that do support membership equality, such as Multiset and Table. We just won't make a defensive copy of those, but the SharePoint connector was the only one doing that, anyway.)

src/com/google/enterprise/adaptor/testing/RecordingDocIdPusher.java
wiarlawd: I have two followups for this CL, one to add javadoc, and the other to propose some semantic changes from what we have in the library and connectors at the moment. I don't know why all of these had a reset method, but I'll look at that. I think there are tests that don't use reset or get a new DocIdPusher, and the assertions look funny, since they include previously collected and asserted results as well.

PapaJoltOmega (approved):

PR #56, opened 2017-05-23
Remove the unused j-calais dependencies.

wiarlawd:
These were all added in commit a7ad120d (Jul 2011)
Related source code removed in commit f80c791 (Jun 2013)
j-calais.jar itself removed in commit 1565b8a (Apr 2015)

No other references to these classes exist, and nothing else depends
indirectly on them. The latter is easy to tell because these JAR files
were referenced only by the examples, and they only import classes
from the JDK and the adaptor library itself.

Remove the otherwise unused licenses from THIRDPARTYLICENSE.txt.


PapaJoltOmega (approved):
Thank you for removing cruft!

PR #64, opened 2017-06-26
Log exceptions thrown out of getDocContent

BrettMichaelJohnson:
Exceptions thrown out of getDocContent weren't getting logged, although they would show up on the console.


wiarlawd (commented):
As discussed, I think these are already logged by HttpServer (a thing you should test), and the logging problem in b/62673598 is in the SharePoint connector itself.

BrettMichaelJohnson:
These are indeed logged in HttpServer.  Abandoning this pull request.

PR #66, opened 2017-08-14
PrebuiltTransformsTest independent of hash order

PapaJoltOmega:


PapaJoltOmega (commented):
Maybe better impl would be to use collection with guaranted sort order in the PrebuiltTransforms themselves.

BrettMichaelJohnson:
All the transforms (not just Delete) end up with a List of processed configuration items, and the order of those lists is dependent on the order of the entries (or keyset) of the config Map passed into the transform. This config map originates in Config.getValuesWithPrefix(), which should probably use either a LinkedHashMap (if that would preserve the order of config in the properties file), or a TreeMap.

BrettMichaelJohnson:
Alternatively, the PrebuiltTransforms all use helper methods to build the Lists, and those helpers could dump the config into a TreeMap before processing.

BrettMichaelJohnson:
Looking at PrebuiltTransformsTest, all the tests create their own config HashMap to pass into the transforms.  It should probably be sufficient just to use TreeMaps in the tests, and leave Config, as-is.

BrettMichaelJohnson:
Ugh! I meant LinkedHashMaps in the tests, as this will preserve the order of the config settings (probably the expected order of the results).

PapaJoltOmega:
Thanks.  Brett, please take another look.

test/com/google/enterprise/adaptor/prebuilt/PrebuiltTransformsTest.java
```diff
@@ -69,13 +70,13 @@ public void testCopyMetadataOverwriteFalse() {
     metadata.add("author", "Fred");
     metadata.add("contributors", "Mary");
     metadata.add("contributors", "George");
-    transform.transform(metadata, new HashMap<String, String>());
+    transform.transform(metadata, new LinkedHashMap<String, String>());
```
BrettMichaelJohnson: I would have only made the configs LinkedHashMaps, and left the rest.

PapaJoltOmega: There are 153 lines with Map on them in this file.  In your opinion is the line-specific thinking worth it? I  did a file replace. Thanks

BrettMichaelJohnson: In general, we perfer to minimize diffs, and the metadata and params maps, and their expected results (which are asserted with Map.equals), don't need to be ordered. In fact it is probably better if they aren't.  It is probably safe to have have done:

s/config = new HashMap/config = new LinkedHashMap/g

PapaJoltOmega: done

test/com/google/enterprise/adaptor/prebuilt/PrebuiltTransformsTest.java
```diff
@@ -706,20 +708,23 @@ public void testDeleteToString() {
...
+    Map<String, String> config = new LinkedHashMap<String, String>();
     config.put("key2", "colour");
     config.put("key3", "author");
     config.put("keyset3", "params");
     config = Collections.unmodifiableMap(config);
 
     MetadataTransform transform = PrebuiltTransforms.deleteMetadata(config);
-    assertEquals("DeleteTransform(keys=[(key=colour,keyset=metadata), "
-        + "(key=author,keyset=params)])", transform.toString());
+    String first = "(key=colour,keyset=metadata)";
```
BrettMichaelJohnson: You should now be able to revert this change.

src/com/google/enterprise/adaptor/Config.java
```diff
@@ -1014,7 +1014,7 @@ File getConfigFile() {
    * @return map of all endings of prefix split into key value pairs
    */
   public synchronized Map<String, String> getValuesWithPrefix(String prefix) {
-    Map<String, String> values = new HashMap<String, String>();
+    Map<String, String> values = new TreeMap<String, String>();
```
BrettMichaelJohnson: This change now become superfluous.

PapaJoltOmega: I was thinking to try to avoid similar future problems.  Is there any benefit to leaving HashMap?

test/com/google/enterprise/adaptor/prebuilt/PrebuiltTransformsTest.java
PapaJoltOmega: is there a reason to do so?

BrettMichaelJohnson: It is no longer necessary, when using the LinkedHashMap, and makes it needlessly complicated to read.

src/com/google/enterprise/adaptor/Config.java
BrettMichaelJohnson: I see no benefit in the change, and even detriment to unnecessary code changes.

test/com/google/enterprise/adaptor/prebuilt/PrebuiltTransformsTest.java
PapaJoltOmega: done

i agree that not necessary. i disagree that complicated or more complicated to read. i liked that it was helpful in debugging.  anyway: done.

src/com/google/enterprise/adaptor/Config.java
PapaJoltOmega: done. i am not sure if beneficial or not. i was thinking to proactively avoid sort problem.  in general i try to use TreeMaps rather than HashMaps for Maps that are typically less than several hundred keys and also for tests. sometimes i forget and type HashMap out of  habit. anyway: done.

test/com/google/enterprise/adaptor/prebuilt/PrebuiltTransformsTest.java
```diff
@@ -706,20 +708,21 @@ public void testDeleteToString() {
...
+    Map<String, String> config = new LinkedHashMap<String, String>();
     config.put("key2", "colour");
     config.put("key3", "author");
     config.put("keyset3", "params");
     config = Collections.unmodifiableMap(config);
 
     MetadataTransform transform = PrebuiltTransforms.deleteMetadata(config);
-    assertEquals("DeleteTransform(keys=[(key=colour,keyset=metadata), "
-        + "(key=author,keyset=params)])", transform.toString());
+    assertEquals("DeleteTransform("
```
BrettMichaelJohnson, line 718: was there a reason to rewrap these lines?

PapaJoltOmega: Minor preference for having the two pieces that need to be in order on the same line, in order.  As dumb as it is I looked multiple times before i realized the actual was out of order from expected. I think having on one line would have helped me a small amount.  It seem that this part of the string is the key part to see.

BrettMichaelJohnson: I accept this exlpanation and this comment..

BrettMichaelJohnson (approved):

wiarlawd:
This was squashed and merged manually as commit f2b47ab. I'm not sure why GitHub didn't notice that, but I'm going to close this PR and delete the branch.

PR #67, opened 2017-08-14
Fix b/62673598 Allow Concurrent Group Pushes

BrettMichaelJohnson:
Removes a restriction by the Journal that prevents concurrent
group pushes, yet still maintains global Journal status that
is reflected in the Dashboard.

Unrelated changes:
- Check completion status on some existing group push tests.
- Add a test for a failed group push.

### DIFF
--- a/src/com/google/enterprise/adaptor/Journal.java
+++ b/src/com/google/enterprise/adaptor/Journal.java
@@ -77,7 +77,17 @@ class Journal {
   private long currentGroupPushStart;
   private long lastSuccessfulGroupPushStart;
   private long lastSuccessfulGroupPushEnd;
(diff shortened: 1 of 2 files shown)


src/com/google/enterprise/adaptor/Journal.java
```diff
@@ -77,7 +77,17 @@
   private long currentGroupPushStart;
   private long lastSuccessfulGroupPushStart;
   private long lastSuccessfulGroupPushEnd;
-  private CompletionStatus lastGroupPushStatus = CompletionStatus.SUCCESS;
+
+  private class GroupPush {
```
PapaJoltOmega: i think could be static.

PapaJoltOmega (approved):
Thank you.

wiarlawd (approved):

PR #72, opened 2017-09-09
Extract DocumentRequest from DocumentHandler.

wiarlawd:
This is intended as a replacement for the various mock Request
implementations in the library and connectors, as well as remaining
the production implementation.

There are a few modest changes to the class:

* Rename to DocRequest for consistency with public names.
* Avoid using HttpExchange in the fields and constructor args.
* Reorder the methods.
* Simplify canRespondWithNoContent.

### DIFF
--- a/src/com/google/enterprise/adaptor/DocRequest.java
+++ b/src/com/google/enterprise/adaptor/DocRequest.java
@@ -35,10 +35,11 @@ public class DocRequest implements Request {
 
   /**
    * Constructs a request with a null access time that supports HTTP
-   * 204 responses. A null access times implies that the document has
+   * 204 responses. A null access time implies that the document has
    * never been crawled.
    *
    * @param docId the requested document ID
+   * @throws NullPointerException if {@code docId} is null
    */
   public DocRequest(DocId docId) {
     this(docId, null);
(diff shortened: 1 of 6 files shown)


src/com/google/enterprise/adaptor/DocRequest.java
```diff
@@ -0,0 +1,113 @@
...
+   * @param lastAccessTime the last time the document was crawled
+   * @param isNoContentSupported {@code true} if an HTTP 204 response
+   *     is allowed, and {@code false} otherwise
+   */
+  public DocRequest(DocId docId, Date lastAccessTime,
+      boolean isNoContentSupported) {
+    this.docId = docId;
+    this.lastAccessTime = lastAccessTime;
+    this.isNoContentSupported = isNoContentSupported;
+  }
```
PapaJoltOmega: what does null doc id mean?

PapaJoltOmega (approved):
Cool.  Thanks.

BrettMichaelJohnson (approved):

src/com/google/enterprise/adaptor/DocRequest.java
wiarlawd: GIGO. The library never passes in null, nor do our connectors, and this class does not (currently) deference it. I'll add a check in the constructor and throw a NPE.

wiarlawd:
Once I added the NPE, I decided to add a test, and once I added one test, I added them for everything. PTAL.

PapaJoltOmega (approved):

PR #75, opened 2017-09-23
Support multiple Java versions with varying support levels.

wiarlawd:


BrettMichaelJohnson (approved):

wiarlawd:
Ping for PJ.

PapaJoltOmega (commented):
What does partial support mean?  Is there a particular feature that is missing with partial support?

Thank you

wiarlawd:
I was trying not to hard-code specifics here, since they are mentioned in the doc and I don't remember the issue well, but it's about Diffie-Hellman and 2048-bit encryption.

https://static.googleusercontent.com/media/www.google.com/en//support/enterprise/static/gsa/docs/admin/connectors/40/413/AdministrationGuideforGoogleConnectors.pdf#page=16

resources/com/google/enterprise/adaptor/TranslationStrings.properties
```diff
@@ -63,5 +63,7 @@ STATUS_JAVA_VERSION=Java Version
 STATUS_JAVA_VERSION_SUPPORTED=Version {0} of Java is supported.
 # The current JVM is not known (i.e. not a standardly-named JVM)
 STATUS_JAVA_VERSION_UNKNOWN=Version {0} of Java is not recognized.  Please ensure it is version {1} (or later).
+# The current JVM is partially unsupported
+STATUS_JAVA_VERSION_PARTIAL=Version {0} of Java is partially supported.  Please upgrade to version {1} (or later).
```
PapaJoltOmega, line 67: This message doesn't communicate the seriousness of the situation. That is, how is admin to judge whether this is serious or trivial.  I think a specific message, and if we can only when we know conditions are met, would be more meaningful and actionable.  Is it possible to detect the condition so as to not scare unnecessarily?

src/com/google/enterprise/adaptor/Dashboard.java
```diff
@@ -252,6 +253,19 @@ public Object run(List request) {
...
   static class JavaVersionStatusSource implements StatusSource {
+    /** Support levels for different Java versions. */
+    private enum Supported { NO, UNKNOWN, PARTIAL, YES };
+
+    /** An ordered map of Java versions to support levels. */
+    private static final Map<String, Supported> versions
+        = new LinkedHashMap<String, Supported>();
+
+    static {
+      versions.put("1.7.0_9", Supported.PARTIAL);
```
PapaJoltOmega, line 264: The doc says:

"""
 If you want to use the DH (Diffie-Hellman) style of encryption and you are running the GSA
with 2048-bit encryption, minimum JRE 1.7u80 or 1.8.0_20 is required.
"""

How large of a population uses DH and runs GSA with 2048 encryption? Can we detect that and give warning then and only then?

PapaJoltOmega (commented):
Thank you.  I have misgivings that this tradeoff is a good one.

resources/com/google/enterprise/adaptor/TranslationStrings.properties
wiarlawd: Rather than adding yet more code, which I don't have any idea how to write or whether that's possible, what if we just replace the second sentence, "Please upgrade ...", to "Please check the supported Java versions in the Administration Guide for details."? This is in all cases a diagnostic message in the logs and dashboard, not a failure.

PapaJoltOmega: Why say "partial"?  Why not say that the version of Java has security problem with Diffie–Hellman?

wiarlawd:
I finally found a good record of this requirement in b/27844124. I think the mention of Java 7u80 is wrong, and I don't know the source for mentioning Java 8u20, but I think this entire CL is unnecessary. Too bad, I kind of liked the code. ;-)

wiarlawd:
Replaced by commit bb2a000.

PR #76, opened 2017-09-26
Allow empty group definitions map.

BrettMichaelJohnson:


PapaJoltOmega (commented):
This CL is not about a group being empty, but about not sending any groups at all, yes?  

Does this behaviour only make sense in "full" push mode?

Thank you

BrettMichaelJohnson:
Yes, it is for an empty Map of group definitions. The connector should be able to nuke all the group definitions by sending an empty full feed, but that was short circuited by the check for isEmpty().  I could have still returned early for incremental feeds, but that required doing so after we may have flipped a full feed to an incremental feed because of GSA version.  An empty group definition map will not actually send a feed to the GSA, because we will never enter the while loop, so it seemed cleaner to just fall through to the end.

PapaJoltOmega:
> " An empty group definition map will not actually send a feed to the GSA"

If feed file is still not sent with this change, then how does a delete by empty file happen?

Thank you

BrettMichaelJohnson:
All the groups are deleted by the empty full feed that sent at the end of every FULL push, and no new groups were uploaded.

PapaJoltOmega:
How do we manage to send "empty full feed that sent at the end of every FULL push" when " An empty group definition map will not actually send a feed to the GSA"?  Is there an alterante mechanism by which the empty full feed is sent?

BrettMichaelJohnson:
The alternate mechanism is on line 421 of DocIdSender.

src/com/google/enterprise/adaptor/DocIdSender.java
```diff
@@ -330,7 +330,7 @@ private String alternateGroupSource(String source) {
       log.log(Level.WARNING,
           "GSA ver {0} does not accept group definitions", gsaVerString);
       journal.recordGroupPushFailed();
-      return defs.entrySet().iterator().next().getKey();
+      return defs.isEmpty() ? null : defs.entrySet().iterator().next().getKey();
```
wiarlawd: Add tests for these two bugs. Maybe change this to defs.keySet().iterator().next(), since we're changing this line anyway?

src/com/google/enterprise/adaptor/DocIdSender.java
```diff
@@ -461,7 +461,7 @@ private String alternateGroupSource(String source) {
       fileArchiver.saveFeed(feedSourceName, groupsDefXml);
       journal.recordGroupPush(defs);
     } else {
-      last = defs.get(0).getKey();  // checked in pushGroupDefinitionsInternal()
+      last = defs.isEmpty() ? null : defs.get(0).getKey();
```
wiarlawd: Setting last to null here mimics success and only works in pushGroupDefinitionsInternal because the batch is empty. I think that's worth a comment.

test/com/google/enterprise/adaptor/DocIdSenderTest.java
```diff
@@ -727,6 +728,70 @@ public void sendGroups(String feedSource, String feedType,
...
+
+    config.setValue("gsa.version", "7.4.0-1");
+    docIdSender = new DocIdSender(fileMaker, fileSender, fileArchiver, journal,
+        config, adaptor);
+
+    // The emptyGroups will not trigger the failure condition in
+    // pushSizedBatchOfGroups() that I'm testing, but the empty full feed
+    // intended to remove previous group definitions will. The return
+    // from pushSizedBatchOfGroups() will be null, simulating SUCCESS.
+    try {
```
wiarlawd: What's the point of the try statements?

BrettMichaelJohnson: Copy/paste from existing test.

wiarlawd (approved):

PapaJoltOmega (approved):

PR #77, opened 2017-09-26
Detect a wrong JDK version in build.bootclasspath.

wiarlawd:


build.xml
```diff
@@ -6,6 +6,7 @@
   <property file="build.properties"/>
   <property name="compile.java.version" value="1.6"/>
   <property name="compile.java.bootclasspath" value="${build.bootclasspath}"/>
+  <property name="compile.java.newerClassname" value="java.lang.AutoCloseable"/>
```
PapaJoltOmega: I don't understand what AutoCloseable has to do with detecting JDK versions. Please annotate, as this looks to be using indirect properties of some aspect rather than something clearly related to JDK versions.

build.xml
```diff
@@ -121,6 +122,12 @@
   <target name="-real-build">
     <mkdir dir="${build-src.dir}"/>
 
+    <available property="compile.java.isNewer"
+      classname="${compile.java.newerClassname}"
+      classpath="${compile.java.bootclasspath}" ignoresystemclasses="true"/>
+    <fail if="compile.java.isNewer"
+      message="build.bootclasspath is newer than JDK ${compile.java.version}."/>
```
PapaJoltOmega: Is this condition bad?  Is this condition good? What can someone do about it?

PapaJoltOmega (commented):
Thank you

build.xml
wiarlawd: It's a class in a newer version of the JDK than 1.6, given two lines above.

build.xml
wiarlawd: Yeah, it's bad, hence the build fail. I'm trying to fix this (documented) error:

If you specify the wrong JDK version in build.bootclasspath, the potential errors are very difficult to detect. The class file versions will be correct, but the code may not run reliably on the desired version of the JVM, and the failures may happen at any time.

build.xml
```diff
@@ -6,6 +6,8 @@
   <property file="build.properties"/>
   <property name="compile.java.version" value="1.6"/>
   <property name="compile.java.bootclasspath" value="${build.bootclasspath}"/>
+  <!-- A class name from the release after the desired version given above. -->
```
PapaJoltOmega: /the release/a release/

Maybe add "This class will NOT be found".

build.xml
```diff
@@ -121,6 +123,14 @@
   <target name="-real-build">
     <mkdir dir="${build-src.dir}"/>
 
+    <!-- Make sure that the bootclasspath isn't newer than expected. -->
+    <available property="compile.java.isNewer"
+      classname="${compile.java.newerClassname}"
+      classpath="${compile.java.bootclasspath}" ignoresystemclasses="true"/>
+    <fail if="compile.java.isNewer">
+      Error: build.bootclasspath is newer than JDK ${compile.java.version}.
```
PapaJoltOmega, line 136: I don't know what this situation means. If the source code compiles with JDK 6 it will also compile with JDK 8.  Are we trying to prevent people from using a new JDK?

build.xml
wiarlawd: It really should be the next release. I was trying not to waste another line of comment on such a minor thing, but I'll expand on this.

build.xml
wiarlawd: This is the thing where the target and bootclasspath must be set together. Setting them separately is error prone. There are several situations handled by this check:

1. If you compile with a later JDK, and build.bootclasspath is not set, then javac will issue a warning to that effect, since the target is set. You'll get a JAR file that may or may not work on the targeted JRE, but which will always work on the build JDK.

2. If you compile with a later JDK, and set build.bootclasspath to point to Java classes of the correct version, then you'll get a JAR file that is guaranteed to work on the target JVM. This is what we want.

3. If you compile with a later JDK, and set build.bootclasspath to point to Java classes also from a later JDK, this check will stop the build, rather than produce a JAR file that may not work on the target JVM. javac doesn't detect this case, but it's the same problem as case 1. I made it an error because we really don't want to build like this.

build.xml
PapaJoltOmega: Yes, please.  

This property name is only used once.  Maybe it should be next to its usage.  One thing that would be alleviated by putting closer to usage is the name "newerClassname" does not convey intent sufficiently all alone ontop of file.  Otherwise a fuller comment, or a really descriptive, and possibly very lengthy name.

build.xml
```diff
@@ -121,6 +123,14 @@
   <target name="-real-build">
     <mkdir dir="${build-src.dir}"/>
 
+    <!-- Make sure that the bootclasspath isn't newer than expected. -->
```
PapaJoltOmega: If someone has only JDK 8 on their machine, they set JAVA_HOME, and then they run ant (without setting more shell variables or ant properties) will compilation proceed for that individual?

PapaJoltOmega (commented):
Thank you

build.xml
wiarlawd: This is case 1. They will get the javac warning, but the build will succeed.

warning: [options] bootstrap class path not set in conjunction with -source 1.6

build.xml
wiarlawd: I explicitly do not want it where it's used, because the usage is boilerplate and does not need to adjusted, and the correct value for this property depends on the value of compile.java.version, set two lines above.

build.xml
PapaJoltOmega: OK.  I think I follow.  Question to double check: Why not treat case #3 like case #1.  In both cases the bootclasspath is not the target JRE bootclasspath.

Another follow up, in order to do things safely, someone needs to have say both JDK8 and JDK6 (or at least JDK8 and JRE6)?

Final followup, the way I do things while developing, I'm in danger case #1. The way I realease things by saying JAVA_HOME=$JDK6 I am in safe case for release. Yes?

build.xml
PapaJoltOmega: OK. 

How about s/newerClassname/classShouldNotExist/ ?

How about much richer comment spelling out this lore:

This build file targets specific JVM version. The below class does not exist in targeted JVM version. It does exist in later versions. We use this class to check provided bootclasspath jars are of correct version.

build.xml
PapaJoltOmega: +1 for adding the entire github review comment you wrote above with the three cases into the build.xml file.

PapaJoltOmega (approved):
Thank you

build.xml
wiarlawd: I made case 3 an error instead of a warning because the person explicitly tried to do the correct thing but failed. I can make it a warning if you prefer.

You just need the classes themselves, and not the whole JRE. On my MacBooks over the years, upgrading Java would sometimes uninstall the old version(s), so I keep the JAR files from previous releases around. I currently only have JDK 8 installed, for example.

In Java 9, all this machinery (-source, -target, -bootclasspath, and needing the older classes) has been replaced by -release. We could, instead, require production builds to use Java 9, and use -release, and if you're not using Java 9, you get whatever random thing your environment produces. Today, we require production builds to set build.bootclasspath to the right JAR files, and if you don't, you get whatever random thing you get.

Yes, you are safe. However, if you set build.bootclasspath in the first place, then it doesn't ever matter what you compile with, invalid code can't sneak into the code base, and you have the option of avoiding compiler bugs.

How would you feel about putting the expanded explanations into the commit message instead?

build.xml
wiarlawd: I had tried changing it to unavailableClassname, but I think that makes the actual code in -real-build look funny. I stuck with newerClassname and an expanded explanation.

wiarlawd (commented):
My latest changes had overlapped your latest comments.

PR #80, opened 2017-10-28
Switch from asList to ImmutableList.of in DodIdSenderTest.

wiarlawd:
Part of removing the suppression of rawtypes warnings.

This change started by avoiding the creation of explicit arrays to
pass to the Arrays.asList varargs method, to reduce clutter and make
it easier to reason about the many nested collections in this class.

Two calls to asList use parameterized types, which results in a
warning about unchecked generic array creation. ImmutableList.of
happens to have overloads with multiple arguments rather than arrays.

This class already used both asList (25%) and ImmutableList.of (75%).
Since I am already editing nearly every call to asList to remove the
explicit arrays, switching the class to consistently use
ImmutableList.of seems reasonable.

### DIFF
--- a/test/com/google/enterprise/adaptor/DocIdSenderTest.java
+++ b/test/com/google/enterprise/adaptor/DocIdSenderTest.java
@@ -35,7 +35,6 @@ import java.io.IOException;
 import java.nio.charset.Charset;
 import java.util.AbstractMap.SimpleImmutableEntry;
 import java.util.ArrayList;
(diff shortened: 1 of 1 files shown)


BrettMichaelJohnson (approved):

BrettMichaelJohnson:
Don't forget to fix the typo in the commit title.

PapaJoltOmega (approved):
Thank you

PR #82, opened 2017-11-01
Remove the suppression of rawtypes warnings.

wiarlawd:
There were five cases:

* Code that expected generic Objects (use a wildcard type parameter)
* Unnecessary type parameters on a class (remove parameterization)
* Unnecessary type parameters on a method (use inferred type)
* Incomplete type parameters on a method (add full type parameters)
* Overrides of API methods that do not use generics (suppress warning)

### DIFF
--- a/src/com/google/enterprise/adaptor/CronScheduler.java
+++ b/src/com/google/enterprise/adaptor/CronScheduler.java
@@ -72,7 +72,7 @@ class CronScheduler {
     if (!(future instanceof CronFuture)) {
       throw new IllegalArgumentException("Provided future is not a CronFuture");
     }
-    CronFuture cronFuture = (CronFuture) future;
+    CronFuture<?> cronFuture = (CronFuture) future;
     CronPattern compiledPattern = CronPattern.create(pattern);
     cronFuture.getCronFilterRunnable().setPattern(compiledPattern);
   }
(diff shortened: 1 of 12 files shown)


src/com/google/enterprise/adaptor/CronScheduler.java
```diff
@@ -72,7 +72,7 @@ public void reschedule(Future<?> future, String pattern) {
     if (!(future instanceof CronFuture)) {
       throw new IllegalArgumentException("Provided future is not a CronFuture");
     }
-    CronFuture cronFuture = (CronFuture) future;
+    CronFuture<?> cronFuture = (CronFuture) future;
```
PapaJoltOmega, line 75: The updated line adds "<?>".  

This addition seems to make less readable to me.  

What do you think?

PapaJoltOmega (commented):
Thank you

src/com/google/enterprise/adaptor/CronScheduler.java
wiarlawd: I think it's the language we have. I really dislike suppressing warnings. It's harmless in cases like this, but not everything that could get suppressed is harmless, and with the warnings suppressed we write code we didn't mean to (e.g., the List<Map.Entry> stuff in DocIdSenderTest).

PapaJoltOmega: What's wrong with List<Map.Entry> in DocIdSenderTest?

PapaJoltOmega (commented):
What's wrong with List<Map.Entry> in DocIdSenderTest?

Thank you

PapaJoltOmega (approved):

PR #83, opened 2017-11-08
Use cleanup feed instead of empty full feed for groups

BrettMichaelJohnson:
It turns out the GSA does not support per-groupsource full feeds.
Any feed with "feedtype=full" replaces all groups from all sources.
But the GSA does support a feed of "cleanup=groupsource" that does
exactly what we need, it removes all group definitions from that
groupsource, so I enhanced GsaFeedFileSender to support cleanup feeds.

Also added "replace=groupsource" support to GsaFeedFileSender since
this file is shared with the GsaFeed project.

### DIFF
--- a/src/com/google/enterprise/adaptor/DocIdSender.java
+++ b/src/com/google/enterprise/adaptor/DocIdSender.java
@@ -418,13 +418,11 @@ class DocIdSender extends AbstractDocIdPusher
         oldGroupSource = previousGroupSource(groupSource);
         groupSources.put(groupSource, feedSourceName);
       }
-      // Delete any entries from the previous group source by pushing an
-      // empty full feed.
+      // Delete any entries from the previous group source by pushing a
+      // cleanup feed.
       log.log(Level.FINE, "Deleting entries from previous group source {0}",
           oldGroupSource);
-      List<Map.Entry<GroupPrincipal, T>> emptyList = Collections.emptyList();
-      pushSizedBatchOfGroups(emptyList, caseSensitive, FULL, oldGroupSource,
-          handler);
+      cleanupGroups(oldGroupSource, handler);
     }
     journal.recordGroupPushSuccessful();
     return null;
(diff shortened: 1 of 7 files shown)


PapaJoltOmega (commented):
Are we leaving "full" feed per GSA meaning in the library?  Or are we replacing the semantics of library's full feed calls with the GSA verb "cleanup"?

src/com/google/enterprise/adaptor/GsaFeedFileSender.java
```diff
@@ -107,9 +109,19 @@ private void buildPostParameter(StringBuilder sb, String name,
   private byte[] buildGroupsXmlMessage(String groupsource, String feedtype,
       String xmlDocument) {
     StringBuilder sb = new StringBuilder();
-    buildPostParameter(sb, "groupsource", "text/plain", groupsource);
-    buildPostParameter(sb, "feedtype", "text/plain", feedtype);
-    buildPostParameter(sb, "data", "text/xml", xmlDocument);
+    String ft = feedtype.toLowerCase(US);
```
wiarlawd, line 116: This isn't ever null, but now it throws a NPE, whereas it previously just sent an invalid parameter to the GSA. Maybe an improvement, but I thought it worth pointing out, and maybe adding an `@throws` tag or `throws` clause for.

src/com/google/enterprise/adaptor/GsaFeedFileSender.java
```diff
@@ -107,9 +109,19 @@ private void buildPostParameter(StringBuilder sb, String name,
...
+    if (ft.equals("full") || ft.equals("incremental")) {
```
wiarlawd: I think that PJ doesn't want to see unused values here ("full" and "replace"), so we'll just address this in gsafeed and reconcile them later if necessary.

test/com/google/enterprise/adaptor/DocIdSenderTest.java
```diff
@@ -450,21 +450,20 @@ public void testPushGroupsReplaceAllGroupsAlternateBuffer() throws Exception {
...
         EVERYTHING_CASE_SENSITIVE, FULL, "foo", null));
 
-    assertEquals(4, fileMaker.i);
+    assertEquals(2, fileMaker.i);
     assertEquals(expectedResult(Integer.MAX_VALUE,
-        sampleGroups(), emptyGroups(), sampleGroups(), emptyGroups()),
+        sampleGroups(), sampleGroups()),
         fileMaker.groupses);
-    assertEquals(ImmutableList.of("incremental", "full", "incremental", "full"),
+    assertEquals(ImmutableList.of("incremental", "cleanup", "incremental", "cleanup"),
```
wiarlawd: 80 cols

src/com/google/enterprise/adaptor/DocIdSender.java
```diff
@@ -471,6 +469,33 @@ private String alternateGroupSource(String source) {
     return last;
   }
 
+  private void cleanupGroups(String feedSourceName,
```
wiarlawd, line 473: This is a fair chunk of replicated code to avoid going through pushSizedBatchOfGroups. What's the win?

wiarlawd (commented):
Minor comments. I'm leaning toward accepting the broken naming, and s/FeedType.FULL/FeedType.REPLACE/.

PapaJoltOmega (commented):
LGTM. Please wait for jlacey@ to approve.

Thank you

wiarlawd (approved):

PR #88, opened 2017-12-10
RecordingResponse.getAnchors() returns a custom multimap.

wiarlawd:
The AnchorMap class is consistent with Guava's LinkedListMultimap.
It has an additional keyList() method to fully support the
LinkedListMultimap semantics of ordered keys and values.

This new functionality is based on our use of the existing,
unreleased getAnchors method in the connectors, and intended to
replace custom, and in some cases incorrect code there.

### DIFF
--- a/src/com/google/enterprise/adaptor/testing/RecordingResponse.java
+++ b/src/com/google/enterprise/adaptor/testing/RecordingResponse.java
@@ -16,6 +16,7 @@ package com.google.enterprise.adaptor.testing;
 
 import static java.util.Collections.unmodifiableList;
 import static java.util.Collections.unmodifiableMap;
+import static java.util.Collections.unmodifiableSet;
 
 import com.google.enterprise.adaptor.Acl;
 import com.google.enterprise.adaptor.Metadata;
@@ -26,11 +27,13 @@ import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.io.OutputStream;
 import java.net.URI;
-import java.util.AbstractMap.SimpleEntry;
+import java.util.AbstractMap.SimpleImmutableEntry;
 import java.util.ArrayList;
 import java.util.Date;
+import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import java.util.TreeMap;
 
 /**
@@ -176,7 +179,7 @@ public class RecordingResponse implements Response {
     if (uri == null) {
       throw new NullPointerException();
     }
-    anchors.add(new SimpleEntry<String, URI>(text, uri));
+    anchors.add(new SimpleImmutableEntry<String, URI>(text, uri));
   }
 
   @Override
@@ -227,7 +230,7 @@ public class RecordingResponse implements Response {
     this.lock = lock;
   }
 
-  // TODO(bmj): @Override
+  @Override
   public void setForcedTransmissionDecision(TransmissionDecision decision) {
     if (state != State.SETUP) {
       throw new IllegalStateException("Already responded " + state);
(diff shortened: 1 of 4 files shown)


BrettMichaelJohnson (approved):

src/com/google/enterprise/adaptor/testing/RecordingResponse.java
```diff
@@ -277,12 +280,103 @@ public boolean isSecure() {
...
+   * An immutable multimap that supports duplicate, ordered key-value pairs.
+   * The anchor texts are the keys, and the anchor URIs are the values.
+   * <p>
+   * This class is consistent with Guava's {@code LinkedListMultimap},
+   * although it does not extend that class, or implement the
+   * {@code ListMultimap} interface or all of its methods.
    */
-  public List<Map.Entry<String, URI>> getAnchors() {
-    return unmodifiableList(anchors);
+  public static class AnchorMap {
```
PapaJoltOmega, line 311: Does it make sense to make this class final? is it possibly correct to subclass it?

src/com/google/enterprise/adaptor/testing/RecordingResponse.java
```diff
@@ -277,12 +280,103 @@ public boolean isSecure() {
...
+    private final List<Map.Entry<String, URI>> anchors;
+
+    private AnchorMap(List<Map.Entry<String, URI>> anchors) {
+      this.anchors = unmodifiableList(anchors);
```
PapaJoltOmega, line 315: do we need a copy?  if someone calls addAnchor after getAnchors their List, that they got before, will change.  Is that OK?

PapaJoltOmega (commented):
Not sure if AnchorMap is as immutable as it should be.  Thank you

src/com/google/enterprise/adaptor/testing/RecordingResponse.java
wiarlawd: The only constructor is private, so it's not possible to subclass it from outside of RecordingResponse.

src/com/google/enterprise/adaptor/testing/RecordingResponse.java
wiarlawd: I think it is, or at least, it's certainly consistent. All of the methods in RecordingResponse that return collections now return unmodifiable views on the underlying data, which can still be changed via the methods of the Response interface. I think we should just make that more clear in the javadoc.

PapaJoltOmega: It's testing code, so being a bit fast-and-loose is fine. 

Yes, there is a difference between "unmodifable" and "immutable" but missing that distinction can catch someone. 

What's the tradeoff in not making copies? Just asking.  Simply not necessary for test code?

Thank you

PapaJoltOmega (approved):
LGTM

Thank you

src/com/google/enterprise/adaptor/testing/RecordingResponse.java
wiarlawd: I don't think we're being fast-and-loose, or writing lesser code because it's aimed at tests. Providing unmodifiable views is different from defensive copies because we're the producer of the data, not the consumer. Making copies would impose an unnecessary decision on the caller. If the caller wants to make sure that the values don't change, then a defensive copy would be reasonable on their side. Our task is simply to make sure we only reflect changes made through the Response interface, and make sure callers don't have direct access to editing the recorded data.

As you suggest, though, with simple, single-threaded tests, the semantic difference is not noticeable.

These classes were an amalgam of half a dozen copies spread through the library (which had more than one) and various connectors. The copies were inconsistent about whether they provided the actual collection references or an unmodifiable view, and I chose the latter when they conflicted. A previous change (#86) made RecordingResponse return unmodifiable views like RecordingDocIdPusher already did, including the anchors.

PapaJoltOmega (approved):
LGTM

Thank you

BrettMichaelJohnson (approved):

PR #90, opened 2018-03-20
Fix PrebuiltTranform.replaceInParams empty and trimmed values

BrettMichaelJohnson:
The configurations for replace string, pattern, and replacement args
are intentionaly not trimmed, allowing patterns, etc, to have leading
and trailing whitespace. Unfortunately PrebuiltTranform.replaceInParams()
would trim the original values read (and perform emptyToNull conversion).
This could cause the value to not match the string or pattern and failed
on empty values.

This change causes replaceInParams() to use the unmodified original values
for the parameters. It also adds more extensive tests for whole value
replacements; empty Strings for string, pattern, replacement, and values;
and untrimmed Strings for string, pattern, replacement, and values.

### DIFF
--- a/test/com/google/enterprise/adaptor/prebuilt/PrebuiltTransformsTest.java
+++ b/test/com/google/enterprise/adaptor/prebuilt/PrebuiltTransformsTest.java
@@ -987,7 +987,9 @@ public class PrebuiltTransformsTest {
     config.put("keyset1", "params");
     config.put("key2", "author");
     config.put("keyset2", "metadata");
-    config.put("pattern", "[ ]+ ");
+    // Pattern with leading and trailing whitespace matches 2 or more spaces,
+    // replacing them with a single space.
+    config.put("pattern", " + ");
     config.put("replacement", " ");
     config = Collections.unmodifiableMap(config);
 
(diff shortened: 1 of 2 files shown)


test/com/google/enterprise/adaptor/prebuilt/PrebuiltTransformsTest.java
```diff
@@ -873,6 +873,187 @@ public void testReplaceStringInParams() {
...
+  }
+
+  @Test
+  public void testReplacePatternUntrimmed() {
+    Map<String, String> config = new LinkedHashMap<String, String>();
+    config.put("key1", "colour");
+    config.put("keyset1", "params");
+    config.put("key2", "author");
+    config.put("keyset2", "metadata");
+    config.put("pattern", "[ ]+ ");
```
wiarlawd: Maybe " + ", plus a comment that you're explicitly testing a pattern with leading and trailing whitespace.

wiarlawd (approved):
LGTM++

test/com/google/enterprise/adaptor/prebuilt/PrebuiltTransformsTest.java
BrettMichaelJohnson: Done.

wiarlawd:
s/Rowlings/Rowling/g

PapaJoltOmega (approved):
Thank you

PR #92, opened 2018-05-16
Ivy dependency handling, full Eclipse support

szab100:
- Auto-download Ivy binary distribution
 - Simplify classpath configuration (compile, runtime, test, misc)
 - Better eclipse support through ant-eclipse ('ant eclipse' => project)
 - Cobertura retrieved by Ivy as well


ivy.xml
```diff
@@ -0,0 +1,34 @@
...
+      <conf name="runtime" description="Additional run-time dependencies"/>
+      <conf name="test" description="Test-only dependencies"/>
+      <conf name="misc" description="Misc components' dependencies (tools, examples)"/>
+      <conf name="buildtools" description="Ant build dependencies"/>
+    </configurations>
+
+    <dependencies>
+        <dependency org="com.googlecode.json-simple" name="json-simple" rev="1.1" conf="compile->default"/>
+        <dependency org="com.google.code.findbugs" name="jsr305" rev="3.0.2" conf="compile->default"/>
+        <dependency org="com.google.guava" name="guava" rev="17.0" conf="compile->default"/> <!-- 17.0 (newest: 25.0-jre) -->
```
szab100: I updated some libraries to newer versions, where it didn't break anything. the comment contains the original versions from lib/ folder. The newest 25.0-jre guava broke the build (easily fixable).

build.xml
```diff
@@ -296,6 +247,7 @@
     <copy todir="${dist.staging.dir}/lib">
       <fileset dir="${lib.dir}" includes="${tmp.adaptorlib.fileset}"/>
     </copy>
+  	<ivy:retrieve conf="compile,runtime" pattern="${dist.staging.dir}/lib/[artifact]-[revision].[ext]"/>
```
wiarlawd: Replace tab.

build.xml
```diff
@@ -1,4 +1,4 @@
-<project name="plexi" default="build" basedir=".">
+<project xmlns:ivy="antlib:org.apache.ivy.ant" name="plexi" default="build" basedir=".">
```
wiarlawd: 80 cols throughout all files.

build.xml
```diff
@@ -31,86 +31,44 @@
...
+  <property name="reports.dir" value="${build.dir}/report"/>
 
   <!-- Use add-modules with Java 9 and higher. -->
   <condition property="java.modules"
       value="" else="--add-modules java.xml.bind">
     <matches string="${ant.java.version}" pattern="^1\.[678]$"/>
   </condition>
 
-  <path id="adaptorlib.build.classpath">
+  <path id="build.classpath">
```
wiarlawd, line 52: These renames seem gratuitous and make me nervous. The library build is called from the connector builds, and even if the clashing names work, they seem open to confusion.

ivy.xml
```diff
@@ -0,0 +1,34 @@
...
+        <dependency org="com.google.code.gson" name="gson" rev="2.8.4" conf="compile->default"/> <!-- 2.0 -->
+        <dependency org="commons-daemon" name="commons-daemon" rev="1.1.0" conf="compile->default"/> <!-- 1.0.15 -->
+
+         <!-- OpenSAML 2 -->
+         <dependency org="org.opensaml" name="opensaml" rev="2.6.6" conf="compile->default"/>
+
+        <!-- Test-only dependencies --> 
+        <dependency org="net.sourceforge.cobertura" name="cobertura" rev="2.1.1" conf="test->default"/>
+        <dependency org="junit" name="junit" rev="4.12" conf="test->default"/> <!-- 4.11 -->
```
wiarlawd: Cobertura includes JUnit 4.11. Even if Ivy manages the dependencies correctly, I wouldn't upgrade unless we really need something from a newer JUnit.

ivy.xml
wiarlawd: When I looked a year ago, Guava 20 worked but was the last version to support Java 6. There's no good reason not to upgrade to 20 (we chose 17 for a reason that no longer has merit).

test/com/google/enterprise/adaptor/SamlIdentityProviderTest.java
```diff
@@ -18,19 +18,17 @@
...
-import org.junit.rules.ExpectedException;
-
 import java.io.IOException;
 import java.util.Arrays;
 import java.util.Set;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
-
 import javax.xml.bind.DatatypeConverter;
+import org.apache.commons.lang.StringEscapeUtils;
```
wiarlawd: We had avoided switching to the new import order piecemeal. Also, we don't have a working lint based on the new scripts, so these are flagged as lint errors at the moment.

build.eclipse.xml
```diff
@@ -1,32 +0,0 @@
...
-  </path>
-
-  <target name="eclipse-build">
-    <javac destdir="${build.dir}" debug="true" encoding="utf-8"
-      includeantruntime="false" target="${compile.java.version}"
-      source="${compile.java.version}">
-      <src path="${src.dir}" />
-      <src path="${test.dir}" />
-      <src path="${resource.dir}" />
-      <compilerarg value="-Xlint:unchecked" />
```
wiarlawd, line 24: These don't match the project's standard lint flags.

build.xml
```diff
@@ -31,86 +31,44 @@
...
-      <include name="commons-logging-api-1.1.1.jar"/>
-    </fileset>
-  </path>
-
-  <path id="cobertura.classpath">
-    <fileset dir="${cobertura.dir}" erroronmissingdir="false">
-        <include name="cobertura.jar"/>
-        <include name="lib/**/*.jar"/>
-    </fileset>
+  <path id="misc.classpath">
```
wiarlawd, line 71: The new class paths are a little less granular. Is that just to make the Ivy config simpler?

test/com/google/enterprise/adaptor/SamlIdentityProviderTest.java
```diff
@@ -96,7 +94,7 @@ public void testNormalFlow() throws Exception {
     ex.getRequestHeaders().set("Host", "bruteforce.mtv.corp.google.com:5678");
     identityProvider.getSingleSignOnHandler().handle(ex);
     assertEquals(200, ex.getResponseCode());
-    String response = new String(ex.getResponseBytes(), "UTF-8");
+    String response = StringEscapeUtils.unescapeHtml(new String(ex.getResponseBytes(), "UTF-8"));
```
wiarlawd, line 101: Is this change related to the updated OpenSAML?

ivysettings.xml
```diff
@@ -0,0 +1,22 @@
+<ivysettings>
+  <!-- Properties to be used in ivy.xml -->
+  <property name="opensaml.version" value="3.3.0" />
```
wiarlawd: This doesn't seem to be used, and ivy.xml refers to version 2.6.6?

wiarlawd (commented):
I like Ivy and upgrading dependencies, but what's the motivation?

build.eclipse.xml
szab100: hmm, well... OK :D At least this is the deletion of this file, since we no longer need this.. :)

build.xml
szab100: Added line breaks, wherever possible, the 'download-ivy' task's long url attribute still exceeds 80.

build.xml
szab100: Indeed. My intention was to simplify the classpaths and reduce the number to compile,runtime,test, misc (for experimental / examples, etc), as it seemed to me that there is no real necessity of having them on a per-component scope..

I can rename them back.. however, can we take a look at the connector builds and see if this could be a problem, so i don't insist to change these, I just thought this makes it simpler.. Since these are internal references, I think it only could be an issue if a connector's ant scrip imports the library's build.xml, which sounds a bit weird to me, but let me check if this is the case..

ivy.xml
szab100: Fair enough.. Let's stick to 4.11.. One additional benefit of using ivy is that it generates a beautiful html report of dependencies, their licenses, transitive deps, conflicts, etc..

build.xml
PapaJoltOmega: Hi szab100,

It looks to me like there are at least two too many intentions behind this CL.  

Smaller CLs are easier to review with higher confidence (less risk).  Smaller CLs can be rolled back individualls. 

I can imagine considering a CL that adds ivey, I can imagine a CL that consider coberatura, I can imagine a CL that considers suggests changing naming of classpaths, i can imagine CL that redoes eclipse.  

Is there really any benefit to doing thise changes together? I think the ivey CL could be OK given that ivey doesn't have to be downloaded seperatly, but what concrete problem are we fixing?

The classpaths are granular, what problem are we fixing?

What is the problem with eclipse?

And what is the problem with cobertura being out? 

I am open to discussing problems or making things better, but I would like us to have articulated objectives for each of the changes. And to have them seperate.

</can chat more later, gotta runt o meeting>

PapaJoltOmega (commented):
Thank you

test/com/google/enterprise/adaptor/SamlIdentityProviderTest.java
szab100: Yes, this newer OpenSAML is html-encoding the post action for security purposes.

ivysettings.xml
szab100: Indeed, removed. Also, to answer your initial question, the motivation behind introducing Ivy was to prepare for OpenSAML 2->3 migration (2.x is deprecated / security-unsupported).

build.xml
szab100: Thank you PJ, yes let's discuss it after your meeting.

1) Smaller CLs are OK, except that it would leave the code broken between the consecutive changelists. However, I wouldn't split it up to more than 2 changelists (Ivy + Eclipse).

2) The classpath names are somewhat linked to and driven by the Ivy change. The general java standard / convention is to define clear and simple dependency & classpath namespaces: compile, runtime, test, (+extras, if necessary). I analyzed the classpath structure and it didn't really make sense to have separate classpath for each components, it added unnecessary complexity and many extra lines to the build file.. So simplifying them brings more clarity in my opinion..

I checked the connectors' sources and none of them imports the library's build file (fortunately, that would be super ugly), so these are internal references only and I would prefer the simplified version. Especially, if we want to introduce Ivy to them as well.

3) Cobertura is still out, meaning we don't include it with the sources. Adding it as an ivy dependency results in extra convenience. Similarly to ivy, users don't have to download it separately. Using it with the 'buildtools' ivy namespace clearly distinguishes it from the code, making it fully gpl-compliant (we can still ask open-source team as @wiarlawd suggested!)

test/com/google/enterprise/adaptor/SamlIdentityProviderTest.java
szab100: restored the original order, hopefully added the new one into the correct place

PapaJoltOmega:
Re: updating secmgr's saml libs
It's likely that updating secmgr component's lib won't hurt.  Perhaps though we should wait until we have either a problem (bug or security issue) with current ones or at least have automated tests or plan for manual testing of adaptors with secmgr
Re: Ivey
Given that Ivey is something that doesn't have to be downloaded by programmer (is that true on every platform? arewe sure?) it might be OK to add it to project. However things are good enough as is so there needs to be a good reason to change them.
Re: Cobertura
Another situation of good enough changing to good enough not being better because it takes time to review and think through.
Re: Renaming/reorganizing classpaths
The current scheme keeps deps small/tailored to target.  It seems like a fine scheme.  I don't see why we'd change from this scheme.
Re: Redoing eclipse tie-in
If you want to propose a seperate CL that redoes how eclipse support is provided then please send CL that just redoes how eclipse is supported in this project, please. If you prefer it for eclipse and it is simply tied into project that seems OK.

I'm going to close this pull request.
Thank you

PR #93, opened 2019-03-30
Add proper Eclipse support through ant-eclipse

szab100:
Remove the old hack with multiple ant build files due to broken / limited eclipse ant plugin support and add proper Eclipse IDE support used by most OSS projects. This plugin creates eclipse project files (.project, .classpath) from ant. All tests pass in Eclipse.

Usage:
 1. Run 'ant eclipse'
 2. Import the project into Eclipse (File / Import / General / Existing Projects into Workspace)


PapaJoltOmega:
What's the javaexec jar file for?  Is there a source that can be included for that? thank you

szab100:
> What's the javaexec jar file for? Is there a source that can be included for that? thank you

That .jar file doesn't contain any class files, only a MANIFEST.MF file with content:

Manifest-Version: 1.0
Main-Class: com.google.enterprise.adaptor.JavaExecTest$Child
Class-Path: ../../../../../build/src/ ../../../../../build/test/ ../../../../../../build/src/ ../../../../../../build/test/

The problem is with class-path entries when running from eclipse vs ant. This is because the resource and java source files are mixed in the test src folder and ant processes only .java files, while Eclipse copies the non-java files too to the output build/test folder. And Eclipse finds and executes this .jar file from the output dir, while ant runs it from the test src folder (explicitly added as resource dir in build.xml for 'test' target). So basically the number of parent dirs ("..") changes.

I realized that this is currently pretty chaotic and usually the correct java project setup is to store java sources separately from non-java resource files. Therefore, to fix this and related, future problems, I added a new commit (rebased to be placed 'before' this ant-eclipse commit) to this PR, which moves the non-java test resource files into its dedicated test-resources folder. (I've done the same previously for production non-java resource files => resources folder).

The JavaExecTestChild.jar file in the 'Move test resources into a dedicated test-resources folder' commit is @ejona86's version from 55aed62df47c1d6ce9fe51719351a4b7d58e3825 checked out, its Manifest.mf contents are simply:

Manifest-Version: 1.0
Main-Class: com.google.enterprise.adaptor.JavaExecTest$Child
Class-Path: ../../../../../build/src/ ../../../../../build/test/

PapaJoltOmega (approved):
LGTM

